PR #102, opened 2014-03-13
initial support for windows.  cygwin and command-prompt

jmckenzie-dev:
I've tried to keep the multiplatform specific support factored out where appropriate.

I've tested this on win8.1, 7, cygwin + msdos for both, and ubuntu 13.10 on 2.0.5 and 1.2.15 with only the issues noted in the README

Note - I'm a perl-guy historically so I welcome all python-idiom critiques.

### DIFF
--- a/README
+++ b/README
@@ -22,8 +22,13 @@ Requirements
     ...
   Note that the usage section assumes that at least 127.0.0.1, 127.0.0.2 and
   127.0.0.3 are available.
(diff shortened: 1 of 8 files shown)


blokhin:
Except *psutil*, works perfectly on my cygwin.
Thanks!
PS one should execute ```powershell Set-ExecutionPolicy Unrestricted```

PR #207, opened 2014-11-07
Added a ccm <node> taillog command

mike-tr-adamson:
This is a simple command that simulates what 'tail -f system.log' does but as a ccm command. It saves having to dive into the .ccm folder to find the log file.


ptnapoleon:
I've removed this temporarily from develop, and soon master, as it is not Python3 compatible. For now, an easy workaround is `ccm node1 showlog | tail`

PR #221, opened 2015-01-18
Add --no-snapshot option to sstablesplit and improve the support to specify files

aboudreault:
This PR contains some improvements for the sstablesplit command:

* Add --no-snapshot option
* Add the capability to specify what file(s) to split. This was broken before and it was splitting all sstables everytime. I fixed this so we can now specify 1 or more files to split.
* This doesn't affect sstable2json, this command doesn't accept sstable file(s) in parameter.
* The only thing that could affect dtests would have been the parameter name change (datafile -> datafiles). I verified and it is not used at all in dtest.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -37,7 +37,7 @@ class TimeoutError(Exception):
         Exception.__init__(self, str(data))
 
 # Groups: 1 = cf, 2 = tmp or none, 3 = suffix (Compacted or Data.db)
-_sstable_regexp = re.compile('(?P<cf>[\S]+)+-(?P<tmp>tmp-)?[\S]+-(?P<suffix>[a-zA-Z.]+)')
+_sstable_regexp = re.compile('(?P<keyspace>[^\s-]+)-(?P<cf>[^\s-]+)-(?P<tmp>tmp(link)?-)?(?P<version>[^\s-]+)-(?P<number>\d+)-(?P<suffix>[a-zA-Z]+)\.[a-zA-Z0-9]+$')
 
 class Node(object):
     """
@@ -745,7 +745,7 @@ class Node(object):
                 shutil.rmtree(full_dir)
                 os.mkdir(full_dir)
 
-    def run_sstable2json(self, out_file=None, keyspace=None, datafile=None, column_families=None, enumerate_keys=False):
+    def run_sstable2json(self, out_file=None, keyspace=None, datafiles=None, column_families=None, enumerate_keys=False):
         print_("running")
         if out_file is None:
             out_file = sys.stdout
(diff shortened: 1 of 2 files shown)


ccmlib/node.py
```diff
@@ -1254,22 +1257,34 @@ def _update_pid(self, process):
...
-                datafiles = datafiles + self.get_sstables(k, "")
-        elif datafile is None:
+            if datafiles:
+                for datafile in datafiles:
+                    if not os.path.isabs(datafile):
+                        datafile = os.path.join(os.getcwd(), datafile)
+
+                    if not os.path.exists(datafile):
+                        print_("File doesn't exist: " + datafile)
+                    elif datafile.endswith('Data.db') or datafile.endswith('big-Data.db'):
```
ptnapoleon: What is the point of this check? Just to ensure only valid files are used? If the file names are coming from the user, I would prefer an exception if they are invalid, rather than silently not appending any files to the call to sstablesplit.

ccmlib/node.py
```diff
@@ -1254,22 +1257,34 @@ def _update_pid(self, process):
...
+    def __gather_sstables(self, datafiles=None, keyspace=None, columnfamilies=None):
+        files = []
         if keyspace is None:
-            for k in self.list_keyspaces():
-                datafiles = datafiles + self.get_sstables(k, "")
-        elif datafile is None:
+            if datafiles:
+                for datafile in datafiles:
+                    if not os.path.isabs(datafile):
+                        datafile = os.path.join(os.getcwd(), datafile)
```
ptnapoleon: Why os.getcwd()? It seems that you're adding the functionality to call sstablesplit on sstables that are not inside the C* data directory here, which is fine, but i don't think that should be in __gather_sstables. I would prefer to isolate the behavior of this function to gathering the paths of sstables inside the C* data directory based on keyspace, CF, etc.

ccmlib/node.py
```diff
@@ -1254,22 +1257,34 @@ def _update_pid(self, process):
...
             if columnfamilies is None:
-                datafiles = datafiles + self.get_sstables(keyspace, "")
+                files = files + self.get_sstables(keyspace, "")
             else:
                 for cf in columnfamilies:
-                    datafiles = datafiles + self.get_sstables(keyspace, cf)
+                    files = files + self.get_sstables(keyspace, cf)
         else:
+            # What is this used for? I don't see where it could be useful... it concatenates the
+            # keyspace dir with a sstable file? Can we remove?
```
ptnapoleon: Can you remove this comment? This is used for when the user passes in the keyspace name, the CF name, and the sstable name. It does appear to be broken, as it is not considering the fact that the CF should be in the path. If you could fix this as part of this PR, that would be appreciated.

ccmlib/node.py
aboudreault: The main reason behind this is to create the same behavior than if we called sstablesplit only one time. With a single sstablesplit call, we can pass something like *9998* as parameter and this will result in only 1 sstable split. If we try that with ccm, this will result in 6 sstable splits.

ccmlib/node.py
aboudreault: Changes proposed:

1. To use datafiles, makes Keyspace param (-k) mandatory 
2. To use datafiles, makes CF param (-c) mandatory 
3. Since I want to support absolute and relative paths matching though bash... an exception will be thrown if the one of the datafiles specified is NOT in the keyspace + column family directory specified in command line.

Specifying a datafile without keyspace and cf will result in the same behavior than before.

ccmlib/node.py
aboudreault: thinking more about that, this will not have the same behavior if the user call the split with in example \*Filter.db,  Which is valid directly through sstablesplit. The best option would be to track the sstable number and only add one file for each. Let me know if this sounds like a reasonable approach.

ptnapoleon: That sounds more reasonable. I hadn't even considered users using wildcards.

ccmlib/node.py
ptnapoleon: +1

ccmlib/node.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, data):
         Exception.__init__(self, str(data))
 
 # Groups: 1 = cf, 2 = tmp or none, 3 = suffix (Compacted or Data.db)
-_sstable_regexp = re.compile('(?P<cf>[\S]+)+-(?P<tmp>tmp-)?[\S]+-(?P<suffix>[a-zA-Z.]+)')
+_sstable_regexp = re.compile('(?P<keyspace>[^\s-]+)-(?P<cf>[^\s-]+)-(?P<tmp>tmp(link)?-)?(?P<version>[^\s-]+)-(?P<number>\d+)-(?P<suffix>[a-zA-Z]+)\.[a-zA-Z0-9]+$')
```
aboudreault: wasn't used anywhere...

ccmlib/node.py
aboudreault: completely removed. The new code handle this case properly.

aboudreault:
Addressed comments and rebased.

ccmlib/node.py
ptnapoleon: Now that you're using this, is the regex also correct for C* 1.2 and 2.0?

ccmlib/node.py
```diff
@@ -1254,22 +1260,43 @@ def _update_pid(self, process):
...
-                datafiles = datafiles + self.get_sstables(keyspace, "")
+                files = files + self.get_sstables(keyspace, "")
             else:
                 for cf in columnfamilies:
-                    datafiles = datafiles + self.get_sstables(keyspace, cf)
+                    files = files + self.get_sstables(keyspace, cf)
         else:
-            keyspace_dir = os.path.join(self.get_path(), 'data', keyspace)
-            datafiles = [ os.path.join(keyspace_dir, datafile) ]
+            if not columnfamilies or len(columnfamilies) > 1 :
```
ptnapoleon: This should be validated in the command's validation as well.

ptnapoleon:
Can you raise more specific errors than Exception? Even if nothing in https://docs.python.org/2/library/exceptions.html is to your liking, it is best to throw NodeError so that it is clear the problem is coming from CCM.

ccmlib/node.py
aboudreault: Yes, the filename format is the same

ccmlib/node.py
aboudreault: Will add it there too.

aboudreault:
Alright

aboudreault:
Addressed last comments. Thanks for the review.

PR #228, opened 2015-01-30
stress command: Forward extra kwargs to subprocess.call

aboudreault:
Useful if we expect a stress command to fail, and want to ignore the 1000+ lines of errors.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -860,7 +860,7 @@ class Node(object):
                 files.remove(f)
         return files
 
-    def stress(self, stress_options=[]):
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Does this work? What does the call to node.stress you are attempting look like?

aboudreault:
Yes, sure it works. I simply use the normal way we do with subprocess.call.
```python
with open(os.devnull, 'w') as devnull:
       self.node1.stress(['--num-keys=1000000'], stdout=devnull, stderr=subprocess.STDOUT)
```

PR #236, opened 2015-02-20
Fix invalid file permissions for running sstable related commands

spodkowinski:
I had trouble to execute all tools except run_sstable2json(). So I moved the permission checking code from there into a new method _find_cmd() which is called by all other run_* methods for retrieving the path to the command with fixed permissions.


ptnapoleon:
Thanks for the PR, and this looks good, but can you close this and open a new PR targeted towards the `develop` branch? It's big enough that I'd like it to have some time getting tested before it goes into master.

PR #242, opened 2015-03-03
Add py Extension During Windows Installation

mikefero:
This will make a copy of the ccm script and give it a py extension if Windows installation detected.


ptnapoleon:
Beautiful. Thanks Fero.

Issue #244, opened 2015-03-10
Use `nodetool info` in Node.data_size


mambocab:
The existing implementation just reads the sizes of files on disk. It should use `nodetool` for this instead. More discussion on these dtest issues/PRs:

https://github.com/riptano/cassandra-dtest/issues/190
https://github.com/riptano/cassandra-dtest/issues/182
https://github.com/riptano/cassandra-dtest/issues/184

mambocab:
I have an implementation that's been added to the dtests as a helper method, which I'll improve and make a PR with if there are no objections to this change.

ptnapoleon:
Not only not objected, but desired.

ptnapoleon:
Fixed by #247

PR #245, opened 2015-03-12
uses `nodetool info` to get node's data size

mambocab:
I've deferred the calculation itself to a "private" function for testability.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -4,6 +4,7 @@ from __future__ import with_statement
 from six import print_, iteritems, string_types
 from six.moves import xrange
 
+import decimal
 import errno
 import glob
 import os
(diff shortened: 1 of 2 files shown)


mambocab:
If merged, will fix #244.

ptnapoleon:
Seeing as how we are changing API here, because `data_size()` no longer accepts live_data as an argument, please point this at develop and not master.

Thank you for the test.

mambocab:
I could add a dummy `live_data` parameter.

Also, that `info=None` parameter shouldn't be there, I'll remove that before merge.

ptnapoleon:
It would be more pythonic to add `args` and `kwargs` to filter out old parameters.

mambocab:
Then we'd have to hand-roll error-checking for spurious arguments. I was thinking something like

```
def data_size(live_data=None):
     if live_data is not None:
         warnings.warn("The 'live_data' keyword argument is deprecated.", DeprecationWarning)
    ...
```

Which wouldn't affect the behavior of arbitrary arguments.

But! sending a PR to `develop` is fine too. Whichever is best for ccm.

ptnapoleon:
I would say I at least want this retargeted to `develop`, then I'll roll it up with some other issues into a minor release. I also like your deprecation solution, that is better than `**kwargs`

mambocab:
Closing; #247 is the real deal.

PR #246, opened 2015-03-13
Use nodetool info for data size develop

mambocab:
as discussed in #245.

Presumably we should remove the deprecation warning at some point. Shall we start a major-version milestone list or something?


mambocab:
Nope, wrong branch. My bad.

PR #247, opened 2015-03-13
uses `nodetool info` to get node's data size

mambocab:
as discussed in #245.

Presumably we should remove the deprecation warning at some point. Shall we start a major-version milestone list or something?


ptnapoleon:
@mambocab Let me worry about that. I'll leave it deprecated and drop it in the next major version.

PR #249, opened 2015-03-18
Don't ever poll process on windows 

### LINKED ISSUE #248: Create cmd with start and binary protocol flag returns before binary protocol started
I've been experiencing on driver our integration tests since updated to the latest 2.0.3 version.
The driver test harness executes a create-and-run command and it returns before the binary protocol started.

I've created a python script that simulates what we are doing on the C# driver test harness:
```python
import os, socket, time
os.system("ccm create test -n 1 -s -i 127.0.0. -b -v 2.1.3")
# time.sleep(1)
sock = socket.socket()
sock.connect(("127.0.0.1", 9042))
sock.close()
```

ptnapoleon:
Fixes #248.

This change relies on the major assumption that we never call `watch_log_for` on a node whose pid we do not know. It also hopes the user has the optional `psutil` dependency installed. `psutil` is optional because it is only required for Windows, and installation via `pip` is problematic, so users must install via the .msi on PyPI. This will make log parsing noticeably slower if `psutil` is not installed.

This PR hopes to solve the issue where with C* 2.1+ on Windows, when not running with legacy startup, the batch script that starts C* returns with exit code zero as soon as the Cassandra Daemon has started up. Thus, the log parsing will exit early, unless the sought after log line is found in the small window where the batch file is still alive.

I have noticed no change in dtest behavior on Linux or Windows, C* 2.0 or 2.1 using this PR. Pointing this at master, as it is a bugfix preventing use of CCM on Windows for some users.

@jorgebay Can you confirm you no longer run into your issue with this PR?
@EnigmaCurry for review.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -378,9 +378,13 @@ class Node(object):
                         raise TimeoutError(time.strftime("%d %b %Y %H:%M:%S", time.gmtime()) + " [" + self.name + "] Missing: " + str([e.pattern for e in tofind]) + ":\n" + reads)
 
                 if process:
-                    process.poll()
-                    if process.returncode is not None and process.returncode == 0:
-                        return None
+                    if common.is_win():
+                        if not self.is_running():
+                            return None
+                    else:
+                        process.poll()
+                        if process.returncode == 0:
+                            return None
 
     def watch_log_for_death(self, nodes, from_mark=None, timeout=600):
         """


EnigmaCurry:
There's three places in this method that run process.poll(), only two of which check if it's running on Windows. Is that intended?

ptnapoleon:
The first method call has never had a Windows check. I believe the logic there is correct, even on Windows. Don't stop waiting for the log file until it either exists, or the start script has terminated. It does seem that the timeout should be respected there as well though, but that should be a separate fix.

mambocab:
A couple nits:

- `x == 0` imples `x is not None`, so you can just check `x == 0`.
- The logic would would be clearer to me if the test were positive: `if common.is_win():`

ptnapoleon:
@mambocab I believe that if `x is None,` then `x == 0` won't return False, but instead will throw `AttributeError`, no? 

I can reverse the logic to be positive

mambocab:
Unless I misunderstand you, I think it should be fine:

```
$ python2 -c 'from __future__ import print_function ; print(0 == None) ; print(None == 0)'
False
False                                                                                                                                     
$ python3 -c 'from __future__ import print_function ; print(0 == None) ; print(None == 0)' 
False
False
```

ptnapoleon:
You appear to be right @mambocab. Changed as requested, and rebased.

jorgebay:
I've tested it on Windows and looks good. Thanks @ptnapoleon!

PR #252, opened 2015-03-20
Fix nodetool options which are not passed

fabienrousseau:
By typing the command below, we can see that the keyspace argument is not taken into account (there is a Note on the beginning of the output of the command):
ccm node1 ring myKS

This pull request allows to pass the options of nodetool commands

### DIFF
--- a/ccmlib/cmds/node_cmds.py
+++ b/ccmlib/cmds/node_cmds.py
@@ -224,20 +224,6 @@ class _NodeToolCmd(Cmd):
     def run(self):
         self.node.nodetool(self.nodetool_cmd)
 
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Thanks for the catch on the duplicate class. It was probably a mistake in merging. I do believe the reason the shorthand nodetool commands weren't accepting arguments was simply because they were being provided as shortcuts, e.g. `ccm node1 status` instead of `ccm node1 nodetool status`. If you currently use `ccm node1 nodetool ring myKS` it will use the argument. That said, I don't really see a compelling reason why we shouldn't accept arguments for the shortcuts.

+1

PR #254, opened 2015-03-25
Add a section on Binary distributions to README

ifesdjeen:
Currently README doesn't mention anything Binary distributions, one can find this information only in code.


ptnapoleon:
Thanks!

PR #255, opened 2015-03-25
Always wait for the pidfile to exist, increase wait

thobbs:
As part of CASSANDRA-8996, this fixes some spurious test failures on trunk.  When ``_update_pid()`` was called from ``cluster.start()``, there was no wait for the pidfile to exist.   Additionally, there was a timeout of 15s before, and trunk was taking about 17s to write the pidfile.  I've increased the wait to 30s to give it some breathing room.


ptnapoleon:
+1. Thanks.

PR #256, opened 2015-03-27
Return float and not Decimal from node.data_size()

ptnapoleon:


ptnapoleon:
Broke the API, clients were expecting float to be returned. Causing TypeErrors in dtest.

PR #257, opened 2015-03-28
Adding Windows support for pause/resume

mikefero:
suspend()
Suspend process execution with SIGSTOP signal pre-emptively checking whether PID has been reused. On UNIX this is the same as os.kill(pid, signal.SIGSTOP). On Windows this is done by suspending all process threads execution.

resume()
Resume process execution with SIGCONT signal pre-emptively checking whether PID has been reused. On UNIX this is the same as os.kill(pid, signal.SIGCONT). On Windows this is done by resuming all process threads execution.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -1354,10 +1354,26 @@ class Node(object):
             return None
 
     def pause(self):
-        os.kill(self.pid, signal.SIGSTOP)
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Fantastic. Thanks @mikefero

PR #260, opened 2015-04-02
Return stdout, stderr from nodetool

thobbs:
I accidentally broke this in commit https://github.com/pcmanus/ccm/commit/10e8438c5c42d9fbbb3dcd168191a86f99fa9c70

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -625,6 +625,8 @@ class Node(object):
         if exit_status != 0:
             raise NodetoolError(" ".join(args), exit_status, stdout, stderr)
 
+        return stdout, stderr
+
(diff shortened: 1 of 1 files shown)


thobbs:
By the way, this is responsible for the dtest failures like this:

~~~
  File "/home/thobbs/cassandra-dtest/concurrent_schema_changes_test.py", line 147, in changes_to_different_nodes_test
    self.validate_schema_consistent(node1)
  File "/home/thobbs/cassandra-dtest/concurrent_schema_changes_test.py", line 114, in validate_schema_consistent
    response = node.nodetool('describecluster', True)[0]
TypeError: 'NoneType' object has no attribute '__getitem__'
~~~

PR #264, opened 2015-04-14
`run_cqlsh` improvements

mambocab:
Several improvements here:

- The way `Node.run_cqlsh` split its input led to lots of lines that looked like `';\n'` getting sent to cqlsh, which leads to confusing syntax errors. This only sends `line + ';'` if `line.strip()` is non-empty.
- If the `show_output` kwarg to `run_cqlsh` is truthy, it prints all output. However, it skipped the first 4 lines. This may have been getting around some property of an old cqlsh, but this is no longer valid.
- Finally, this adds a `return_output` kwarg. If it's truthy, `run_cqlsh` will return cqlsh's output as a string.

Since `return_output` is a kwarg with a default value at the end of the argument list, this shouldn't change the `run_cqlsh` API; if you just ignore it, it won't affect you.

 I've also tested the combination of these two kwargs and added a basic "does this thing fall over" test.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -695,7 +695,9 @@ class Node(object):
         else:
             p = subprocess.Popen([cqlsh] + args, env=env, stdin=subprocess.PIPE, stderr=subprocess.PIPE, stdout=subprocess.PIPE)
             for cmd in cmds.split(';'):
-                p.stdin.write(cmd + ';\n')
+                cmd = cmd.strip()
+                if cmd:
+                    p.stdin.write(cmd + ';\n')
             p.stdin.write("quit;\n")
             p.wait()
             for err in p.stderr:
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -703,12 +703,8 @@ class Node(object):
             for err in p.stderr:
                 print_("(EE) ", err, end='')
             if show_output:
-                i = 0
                 for log in p.stdout:
-                    # first four lines are not interesting
-                    if i >= 4:
-                        print_(log, end='')
-                    i = i + 1
+                    print_(log, end='')
 
     def cli(self):
         cdir = self.get_install_dir()
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -1,7 +1,7 @@
 # ccm node
 from __future__ import with_statement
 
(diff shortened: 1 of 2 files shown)


ptnapoleon:
How does this affect the cqlsh dtests?

mambocab:
The semicolon-splitting change doesn't affect behavior other than to get rid of some errors during test-writing that were confusing me. Some dtests avoided this by removing the final line-terminating semicolon ([example](https://github.com/riptano/cassandra-dtest/blob/b471b5cf6c35ca0ee822e217a662711be1d95c9b/cqlsh_tests.py#L346). This workaround won't be necessary anymore. However, with the way the input is executed line by line ([here](https://github.com/pcmanus/ccm/blob/master/ccmlib/node.py#L697)), the line-terminating semicolons are added back after the split, so these workarounded inputs still work. I ran the cqlsh dtests with the changed ccm to confirm.

The change to `show_output`'s behavior shouldn't affect the dtests, since [show_output isn't used there](https://github.com/riptano/cassandra-dtest/search?utf8=%E2%9C%93&q=show_output). It's just a fix so it behaves as expected.

The `return_output` kwarg removes the need to [duplicate the `run_cqlsh` method in the dtests](https://github.com/riptano/cassandra-dtest/blob/b471b5cf6c35ca0ee822e217a662711be1d95c9b/cqlsh_tests.py#L488). It's confusing and unnecessary to have both.

mambocab:
Oof, the logic for getting the output to return it is much more complicated than it needs to be and doesn't return stderr. It should just return the tuple returned by `process.Popen`. @ptnapoleon If you think this is a worthwhile feature, I'd like to force push a better implementation.

ptnapoleon:
I'm fine returning the `stdout, stderr` tuple, if that's what you mean. Feel free to rebase

mambocab:
Better implementation pushed. Thoughts?

ptnapoleon:
That's clever. There are some other commands that could probably be cleaned up in the same way. They return the raw `p.stdout` and `p.stderr`, rather than the result of `read()`.

ptnapoleon:
Dtests still good? If so, I'll merge

mambocab:
dtests still good over here.

PR #266, opened 2015-04-16
Remove all gentle/stop-server.bat logic for stopping nodes on Windows

josh-mckenzie:


ptnapoleon:
-1. Check out def __update_status_win(self). psutil isn't installed/used on linux, so it needs wrapped in a try/except ImportError block, and only if running on windows.

josh-mckenzie:
Missed that we already had an example of that usage (and should have tested this on linux - sorry about that). Promoted that to _find_pid_in_windows() and refactored to use that method in both stop() and update_status_win(). Tested on linux and win, rebased to squash and force-pushed.

PR #272, opened 2015-05-05
Add option for enabling internode encryption.

csm:
* ccmlib/cluster.py (Cluster.enable_internode_ssl): new method.
* ccmlib/cmds/cluster_cmds.py (CreateClusterCmd.get_parser): add `--node-ssl` option.
(CreateClusterCmd.run): set up internode encryption if needed.
* ssl/truststore.jks: new file.


ptnapoleon:
+1. Thanks for the patch!

PR #279, opened 2015-05-07
278

blerer:
The patch make sure that we do not overrides the configuration files in  common.make_cassandra_env and fix the wrong JMX configuration that was made for Windows in node.__update_envfile.


ptnapoleon:
-1. This patch does not work on OSX, and I suspect Linux.

ptnapoleon:
Thanks for the fix.

PR #283, opened 2015-05-12
fix/refactor/test Node.grep_log_for_errors

mambocab:
This is a biggish change, but I think the tests demonstrate it is good.

The problem that spurred this was that DEBUG messages containing the string ERROR would spuriously be treated as errors. I ran a dtest that generated errors, and the grep called on the test teardown successfully made the test fail, so that should still work.


ptnapoleon:
Looks great, +1 and thanks!

PR #289, opened 2015-05-21
Print pid of nodes when started on Windows

josh-mckenzie:
This is an interim change to facilitate tracking down if specific dtests are consistently hanging. Windows only and just a little verbosity on node start.


ptnapoleon:
This won't work on Python 3, sorry. -1

josh-mckenzie:
Updated to use six print version to stderr.

ptnapoleon:
+1! Thanks.

PR #294, opened 2015-06-02
rmdir() to allow rmtree on path > 255 chars on Windows

kishkaru:
Created new helper method common.rmdir() to allow rmtree on path > 255 chars on Windows.


ptnapoleon:
I'm not in love with this fix, but I understand your need for it.

PR #295, opened 2015-06-05
Added support for package level logging

steveandwang:
Modified the replaces_or_add_into_file_tail function so that it adds additional lines before the </configuation> line

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -121,10 +121,12 @@ def replaces_or_add_into_file_tail(file, replacement_list):
                     if match:
                         line = replace + "\n"
                         is_line_found = True
-                f_tmp.write(line)
+                if line != "</configuration>\n":
+                    f_tmp.write(line)
             # In case, entry is not found, and need to be added
             if is_line_found == False:
-                f_tmp.write('\n'+ replace + "\n")
+                f_tmp.write('\n' + replace + "\n")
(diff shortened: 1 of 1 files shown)


ccmlib/common.py
```diff
@@ -121,10 +121,12 @@ def replaces_or_add_into_file_tail(file, replacement_list):
                     if match:
                         line = replace + "\n"
                         is_line_found = True
-                f_tmp.write(line)
+                if line != "</configuration>\n":
```
ptnapoleon: Instead, could you use a less restrictive check? Such as `if "</configuration" not in line`?

ccmlib/common.py
```diff
@@ -121,10 +121,12 @@ def replaces_or_add_into_file_tail(file, replacement_list):
...
+                    f_tmp.write(line)
             # In case, entry is not found, and need to be added
             if is_line_found == False:
-                f_tmp.write('\n'+ replace + "\n")
+                f_tmp.write('\n' + replace + "\n")
+            f_tmp.write("</configuration>\n")
```
ptnapoleon: Could you comment this line so it's clear why we're moving the closing tag?

ptnapoleon:
Doesn't appear to break anything. If you could make the few small changes, and then squash the commits, I'll merge.

ptnapoleon:
Closed in favor of #298.

PR #297, opened 2015-06-08
Added support for package levels

steveandwang:
Modified the replaces_or_add_into_file_tail function so that it adds changes to class levels before the ending tag.


ptnapoleon:
Closing this. Keeping focus in #295

PR #300, opened 2015-06-08
5322

steveandwang:
Updated environment variables DEBUG and TRACE such that they also accept class names along with true/false/yes/no, By inputting class name(s) (separated with a colon), you can set only those classes to have that level of logging. Updated the set_log_level feature so that it wouldn't automatically change the global level, and would update individual class log_level modifications.


ptnapoleon:
This can't be merged. Rebase and squashed your commits onto master.

PR #301, opened 2015-06-09
New sstable filename doesn't contain keyspace and column family

aboudreault:


ptnapoleon:
This still works on older versions, yes? I see the `?` in the regex, but wanted to be sure you checked.

aboudreault:
Yes

PR #302, opened 2015-06-09
5322

steveandwang:
Modified set_log_level to support multiple class names.

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -26,6 +26,9 @@ class Cluster(object):
         self.__path = path
         self.__version = None
         self.use_vnodes = False
+        # Classes that are to follow the respective logging level
+        self._debug = []
+        self._trace = []
 
         ##This is incredibly important for
         ##backwards compatibility.
@@ -112,6 +115,12 @@ class Cluster(object):
         self._update_config()
         node.data_center = data_center
         node.set_log_level(self.__log_level)
+
+        for debug_class in self._debug:
+            node.set_log_level("DEBUG", debug_class)
+        for trace_class in self._trace:
+            node.set_log_level("TRACE", trace_class)
+
         if data_center is not None:
             self.__update_topology_files()
         node._save()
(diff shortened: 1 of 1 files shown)


ccmlib/common.py
```diff
@@ -126,9 +126,7 @@ def replaces_or_add_into_file_tail(file, replacement_list):
             # In case, entry is not found, and need to be added
             if is_line_found == False:
                 f_tmp.write('\n' + replace + "\n")
-            # We are moving the closing tag to the end of the file.
-            # Previously, we were having an issue where new lines we wrote
-            # were appearing after the closing tag, and thus being ignored.
+            # Want to close the file with last line "<configuration>\n"
```
ptnapoleon: You're keeping the wrong comment here. We want the longer one. You can just reset all changes to common.py, since this is the only difference.

ccmlib/cluster.py
```diff
@@ -293,16 +302,28 @@ def stop(self, wait=True, gently=True):
...
+    def set_log_level(self, new_level, class_names=None):
         known_level = [ 'TRACE', 'DEBUG', 'INFO', 'WARN', 'ERROR' ]
         if new_level not in known_level:
             raise common.ArgumentError("Unknown log level %s (use one of %s)" % (new_level, " ".join(known_level)))
 
-        self.__log_level = new_level
-        self._update_config()
+        if class_names:
+            for class_name in class_names: 
+                if new_level is 'DEBUG':
```
ptnapoleon: You shouldn't use `is` here, the proper test for equality is `==`.

ccmlib/cluster.py
```diff
@@ -293,16 +302,28 @@ def stop(self, wait=True, gently=True):
...
+                    if class_name in self._trace:
+                        raise common.ArgumentError("Class %s already in TRACE" % (class_name))
+                    self._debug.append(class_name)
+                if new_level is 'TRACE':
```
ptnapoleon: Ditto

ptnapoleon:
If you could fix those two nits, then +1.

PR #305, opened 2015-06-15
Add 'kill' command to SIGKILL a node.

tolbertam:
For testing, it is useful to be able to send a SIGKILL to a node to have
it uncleanly terminate.  In this case the OS will send TCP RST messages to
clients which may cause them to behave in unpredictable ways.


tolbertam:
I think #45's --not-gently may cover my use case, so this PR may not be needed...testing.

tolbertam:
Looks like --not-gently does the job.   Closing this PR but feel free to reopen if you find 'kill' useful.

Issue #293, opened 2015-05-29
Allow creating clusters from arbitrary repositories


ptnapoleon:
It would be great if when passing in `-v git:BRANCH`, if users could specify an arbitrary repository and branch, as long as it was a functioning C* fork/mirror.

PR #309, opened 2015-06-17
GitHub repos

mambocab:
Allows for creation of clusters from github by specifying `github:username/branch_name` instead of `git:branch_name`, partially addressing #293.

This is pretty quick-and-dirty, but it seems to work. It creates a new `_git_cache` directory for each GitHub repo cloned, which could get big after a while.

This changes the existing format for cluster repo directories. Instead of replacing `:` with `_`, it replaces `:` with `COLON` and `/` with `SLASH`. Some replacement for the latter is necessary since `/` is valid in branch names, so we may encounter them in GitHub branches. Replacing them both with uppercase ascii gives each a unique replacement and makes it unlikely that directory names will collide. Users will want to run `ccm invalidatecache` after updating.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,20 @@ If you'd like to use a binary distribution instead of compiling from sources eac
 ccm create test -v binary:2.0.5 -n 3 -s
 ```
 
+### Git and GitHub
+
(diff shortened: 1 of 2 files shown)


ptnapoleon:
+1, thanks for the README.

PR #312, opened 2015-06-22
add option to block on the log for Node.drain

mambocab:
I don't know if this goes against some design philosophy for `node.Node`, so I'm just running this up the flagpole. Seems like [this kind of logic](https://github.com/riptano/cassandra-dtest/pull/338/files#diff-abac0928e2a06eb362676fa694227510R122) could be repeated all over the place in dtests, so why not consolidate that all into one place?

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -958,8 +958,11 @@ class Node(object):
     def compact(self):
         self.nodetool("compact")
 
-    def drain(self):
+    def drain(self, block_on_log=False):
+        mark = self.mark_log()
         self.nodetool("drain")
+        if block_on_log:
+            self.watch_log_for("DRAINED", from_mark=mark)
 
     def repair(self, options=[], **kwargs):
         args = ["repair"] + options


ptnapoleon:
Does `nodetool drain` not already block? Is it possible we're just being needlessly careful? Also, you need to mark the log before calling drain, and use that when searching, if you want this to be idempotent.

mambocab:
> Does nodetool drain not already block? Is it possible we're just being needlessly careful?

It is entirely possible. I was just assuming based on the use I saw.

I tried to find where the draining actually happens, but the trail went cold after I figured out that this works by [calling `drain` on some kind of `MBeanProxy`](https://github.com/apache/cassandra/blob/03f556ffa8718754fe4eb329af2002d83ffc7147/src/java/org/apache/cassandra/tools/NodeProbe.java#L347). You like Java, right, Philip?

> Also, you need to mark the log before calling drain, and use that when searching, if you want this to be idempotent.

That's true. If the proposed method winds up being something other than a noop, I'll change that.

ptnapoleon:
It does block. So I think we're usually just being extra careful. However, this is no harm in going through with this implementation. So if you add the pre-drain log mark, I'll merge.

mambocab:
Since it already blocks, I renamed the parameter so it's clear that we're blocking on the log. I also added the mark.

ptnapoleon:
+1

PR #316, opened 2015-06-24
Change suggested command on cluster op failure

### LINKED ISSUE #315: Incorrect hint when trying to remove a cluster that doesn't exist
Upon trying to remove a cluster that doesn't exist, ccm says: `[clustername] does not appear to be a valid cluster (use ccm cluster list to view valid cluster)`. The correct command is actually `ccm list` if I'm not mistaken.

jkingsman:
`ccm list` lists valid clusters. `ccm cluster list` is not a valid command. Fixes #315.


ptnapoleon:
+1

PR #318, opened 2015-06-29
makes is_win more readable and idiomatic

mambocab:
This is just a little more compact and easy to read. Jankily tested from the interpreter:

```python
>>> from ccmlib import common
>>> import inspect
>>> inspect.getsourcelines(common.is_win)  # check that I'm using my new implementation
(['def is_win():\n', '    return sys.platform in ("cygwin", "win32")\n'], 215)
>>> # I totally am
...
>>> common.is_win()
False
>>> import sys ; sys.platform = 'cygwin'
>>> common.is_win()
True
```


ptnapoleon:
+1

PR #324, opened 2015-07-06
Changes to support multiple data directories

shawnkumar:
Nothing too crazy - just add a class variable which defaults to the current setup and method to change such variable. Had to change some methods that had been hardcoded to the default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,10 +56,6 @@ There is also a [Homebrew package][brew] available:
 
     brew install ccm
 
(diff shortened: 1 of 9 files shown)


shawnkumar:
Not sure if I messed something up here - should I be committing the changes to develop on my fork?

ptnapoleon:
I'll close this PR. Open a new one targeted at master.

PR #325, opened 2015-07-06
Additions to support multiple data directories

shawnkumar:


ccmlib/node.py
```diff
@@ -96,6 +96,7 @@ def __init__(self, name, cluster, auto_bootstrap, thrift_interface, storage_inte
         self.__install_dir = None
         self.__global_log_level = None
         self.__classes_log_level = {}
+        self.data_dirs = ['data']
```
ptnapoleon: Can you make this variable private?

ptnapoleon:
So the code looks good, and I'm +1, but since this is a pretty core change, can you get a cassci run with your branch and let me know that this doesn't break any dtests, please?

ptnapoleon:
Closing this. Markus has a more thorough implementation

PR #326, opened 2015-07-06
Improvements for updateconf, updatedseconf commands

bcantoni:
Changes:
- Remove leftover references to dse_yaml_file option which is no longer supported
- Explain how to specify option.suboption with period separator
- Fix for delete conf setting by not providing the value (ex: "ccm updateconf key:")

I made the 3rd item a separate commit to make the fix more clear.


ptnapoleon:
+1. Thanks!

PR #335, opened 2015-07-23
modified functions

steveandwang:
Added a couple of utilities to node, will update later with some comments as to how they work.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -957,11 +957,26 @@ class Node(object):
         info = self.nodetool('info', capture_output=True)[0]
         return _get_load_from_info_output(info)
 
-    def flush(self):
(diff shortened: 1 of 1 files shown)


ptnapoleon:
-1 on naming those arguments `wait_for_binary_proto`, since there are not waiting for the binary proto. `block_on_log`, `watch_log`, etc would all be fine. But isn't the nodetool command already blocking? Also, you need to mark the log and then resume from the mark for this to work idempotently.

ptnapoleon:
Also, drop those sleeps. I know why you put them there, but we don't need them.

ptnapoleon:
Please rename that `compaction` command to be more clear as to what it does

ptnapoleon:
Fix the marks, and then I'm +1

PR #336, opened 2015-07-24
Fix bulkload command.

kdeldycke:
`bulkload` command wasn't working so I fixed it and re-integrated it along other commands.


ptnapoleon:
+1, thanks for the PR. I didn't realize this hadn't been removed yet, but just as happy to have it fixed.

kdeldycke:
Wow, that was fast ! Thantks for the merge ! :+1: :smile:

PR #337, opened 2015-07-27
Revert "modified functions"

ptnapoleon:
Reverts pcmanus/ccm#335


ptnapoleon:
This broke many dtests not on trunk. @steveandwang, you'll need to modify the changes you made so they work on all versions.

PR #338, opened 2015-07-28
allow ccm to run with ant installed from chocolatey package manager

pauloricardomg:
for some reason, the chocolatey package manager on windows installs ANT with the executable ant.exe instead of ant.bat, so ccm is not working because it expects an ant.bat executable. a simple fix is to omit the file extension on windows, which is currently hard-coded to .bat.

on windows7+, if the $PATHEXT variable is set, we can safely ommit the executable extensions listed on that variable while running an executable, so for example, we can run just ant instead of ant.bat.

since ".exe" and ".bat" are listed by default on $PATHEXT we can safely omit the ".bat" extension on ccmlib.common.platform_binary for windows hosts, if .bat is listed in this $PATHEXT variable. if this variable is not set, we still add the .bat suffix to windows binaries, so we maintaint backwards compatibility.

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -217,9 +217,9 @@ def check_win_requirements():
     if is_win():
         # Make sure ant.bat is in the path and executable before continuing
         try:
-            process = subprocess.Popen('ant.bat', stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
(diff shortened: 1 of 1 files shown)


pauloricardomg:
closing as this has nasty side-effects

pauloricardomg:
simple fix is to create a symbolic link:
cmd /c mklink C:\ProgramData\chocolatey\bin\ant.bat C:\ProgramData\chocolatey\bin\ant.exe

PR #343, opened 2015-07-31
Minor fixes to support #CASSANDRA-9942

pauloricardomg:
- Added support to "OFF" log level
- Changing log levels on logback-tools.xml
- Exporting CASSANDRA_HOME and CASSANDRA_CONF vars


ptnapoleon:
+1, thanks

Issue #341, opened 2015-07-31
Support clusters running mixed Cassandra versions


spodkowinski:
It would be great to be able to use ccm to test upgrade scenarios by upgrading individual nodes one by one. This would involve stopping a single node using ccm, changing the install dir while keeping the data dir, and starting the node again. This would also be useful for writing dtest cases for updates, as there will be more frequent Cassandra releases in the future.

spodkowinski:
Looks like this seems to be already possible from the api, see *upgrade* tests in cassandra-dtest.

ptnapoleon:
It is technically also possible from the command line, albeit clunky and unpleasant.

spodkowinski:
I've just modified the setversion command to make this more convenient. Will do a PR after I'm done.
https://github.com/pcmanus/ccm/compare/master...spodkowinski:feat/setnodeversion?expand=1

PR #345, opened 2015-08-03
Allow setting directory for individual nodes

spodkowinski:
This PR will add an option to run `setdir` for individual nodes (`*/node.conf`) instead of the whole cluster (`cluster.conf`). As briefly discussed in #341.


ptnapoleon:
Lgtm. Thanks!

PR #346, opened 2015-08-04
Added node.waitforcompactions() method

pauloricardomg:
This method is being used in more than one class of cassandra-dtest so I thought it would be useful to bring this up to ccm.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -580,6 +580,17 @@ class Node(object):
         else:
             return False
 
+    def wait_for_compactions(self):
+        """
+        Wait for all compactions to finish on this node.
+        """
+        pattern = re.compile("pending tasks: 0")
+        while True:
+            output, err = self.nodetool("compactionstats", capture_output=True)
+            if pattern.search(output):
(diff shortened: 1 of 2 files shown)


ccmlib/node.py
```diff
@@ -580,6 +580,17 @@ def stop(self, wait=True, wait_other_notice=False, gently=True):
...
+    def wait_for_compactions(self):
+        """
+        Wait for all compactions to finish on this node.
+        """
+        pattern = re.compile("pending tasks: 0")
+        while True:
+            output, err = self.nodetool("compactionstats", capture_output=True)
+            if pattern.search(output):
+                break
+            time.sleep(30)
```
ptnapoleon: Can you bump this down to 10 or 15 seconds? And can you add a cluster.wait_for_compactions that just calls wait_for_compactions on each node?

ptnapoleon:
+1, but could you make the suggested changes? Thanks!

pauloricardomg:
done, thanks for the suggestions!

PR #349, opened 2015-08-07
Establish socket to binary port for --wait-for-binary-proto to ensure port is listening.

tolbertam:
I was encountering a very rare and small timing window when running tests on a constrained VM (2 cores, 4GB ram) that had ongoing load where the binary interface wasn't available after ccm start even if --wait-for-binary-proto was used.

Replacing `time.sleep(0.2)` with a method that verifies the binary interface is listening by attempting to create a socket to it seems to resolve my issue (so far).   Will try for up to 10 seconds (more than reasonable if 'Starting listening for CQL clients' is spotted in the logs) and prints a warning and continues on if the interface is still not available.

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -291,10 +291,9 @@ class Cluster(object):
                     if old_node is not node:
                         old_node.watch_log_for_alive(node, from_mark=mark)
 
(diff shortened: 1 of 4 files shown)


ptnapoleon:
Looks great, thanks Andy!

PR #353, opened 2015-08-17
Adds creation of hints directory, introduced in CASSANDRA-6230

EnigmaCurry:
This is causing testing of CASSANDRA-6230 to go haywire:

http://cassci.datastax.com/job/iamaleksey-6230-3.0-dtest/4/testReport/junit/junit/auth_test/TestAuth/login_test/

### DIFF
--- a/ccmlib/dse_node.py
+++ b/ccmlib/dse_node.py
@@ -350,7 +350,7 @@ class DseNode(Node):
 
     def _get_directories(self):
         dirs = {}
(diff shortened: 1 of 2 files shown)


knifewine:
Looks pretty innocuous to me. Does dse_node.py need the same change?

+1

EnigmaCurry:
Good catch. It doesn't yet, but it will once DSE pulls in 6230.

EnigmaCurry:
This doesn't create the directories in the correct order, because _get_directories returns a dictionary.

PR #356, opened 2015-08-18
Flush before kill on Windows. Add retry on deletion in cluster.remove()

josh-mckenzie:
Reference CASSANDRA-10075: The retry is an attempt to remedy an intermittent
failure in dtests where commitlogs cannot be deleted during cluster stop.
Flush before kill is an attempt to address a class of problems that has
shown up in various dtests while stabilizing Windows tests for 2.2.

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -209,7 +209,7 @@ class Cluster(object):
 
     def remove(self, node=None):
         if node is not None:
-            if not node.name in self.nodes:
+            if node.name not in self.nodes:
                 return
 
             del self.nodes[node.name]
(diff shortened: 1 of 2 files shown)


ccmlib/cluster.py
```diff
@@ -209,7 +209,7 @@ def balanced_tokens_across_dcs(self, dcs):
 
     def remove(self, node=None):
         if node is not None:
-            if not node.name in self.nodes:
+            if node.name not in self.nodes:
```
ptnapoleon: Oh boy who had that backwards.

ccmlib/node.py
```diff
@@ -561,6 +561,11 @@ def stop(self, wait=True, wait_other_notice=False, gently=True):
                 # We have recurring issues with nodes not stopping / releasing files in the CI
                 # environment so it makes more sense just to murder it hard since there's
                 # really little downside.
+
+                # We want the node to flush its data before shutdown as some tests rely on small writes being present.
+                # The default Periodic sync at 10 ms may not have flushed data yet, causing tests to fail.
+                self.flush()
```
ptnapoleon: Can we possibly filter this on if graceful=True?

ccmlib/cluster.py
josh-mckenzie: flake8 for ccm is... not a pretty picture. I figured "leave the place cleaner than when I got here".

ptnapoleon: Yeah, I'll flake8 it. I already ran some pep8, and isort.

ccmlib/node.py
josh-mckenzie: Force pushed update w/check for if gently is True

ccmlib/node.py
```diff
@@ -561,6 +561,12 @@ def stop(self, wait=True, wait_other_notice=False, gently=True):
                 # We have recurring issues with nodes not stopping / releasing files in the CI
                 # environment so it makes more sense just to murder it hard since there's
                 # really little downside.
+
+                # We want the node to flush its data before shutdown as some tests rely on small writes being present.
+                # The default Periodic sync at 10 ms may not have flushed data yet, causing tests to fail.
+                if gently is True:
```
ptnapoleon: Next time, better just to do `if gently:`

PR #359, opened 2015-08-18
6177 sleep

steveandwang:
Allow flush/compact to be called without a sleep following them.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -828,7 +828,7 @@ class Node(object):
             subprocess.call(cmd, cwd=os.path.join(self.get_install_dir(), 'bin'), env=env, **kwargs)
 
         for sstablefile in sstablefiles:
-            do_split(sstablefile)
+            do_split(sstablefile) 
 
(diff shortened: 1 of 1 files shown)


ccmlib/node.py
```diff
@@ -1011,11 +1011,29 @@ def data_size(self, live_data=None):
...
         return _get_load_from_info_output(info)
 
-    def flush(self):
+    def flush(self, delay=True):
+        # The "delay" argument allows one to determine whether to delay the test until the flush command is finished. 
+        mark = self.mark_log()
         self.nodetool("flush")
+        if delay and self.cluster.version() >= '1.2':
+            try:
+                self.watch_log_for("Completed flushing", from_mark=mark, timeout=10)
```
ptnapoleon: Are you sure this log message shows up in all versions >= 1.2?

ptnapoleon:
You'll definitely at least need to squash and merge. Can you possibly run some tests that use these methods on every version >= 2.0 and let me know that they work?

PR #360, opened 2015-08-19
make flush failure silent on Windows node stop

josh-mckenzie:
Caused 11 new failures since flush failure can rebound w/raised exception. If we fail to flush on shutdown it's not really an error, so changed to swallow exception during stop() on Windows.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -564,8 +564,12 @@ class Node(object):
 
                 # We want the node to flush its data before shutdown as some tests rely on small writes being present.
                 # The default Periodic sync at 10 ms may not have flushed data yet, causing tests to fail.
+                # This is not a hard requirement, however, so we swallow any exceptions this may throw and kill anyway.
                 if gently is True:
-                    self.flush()
+                    try:
+                        self.flush()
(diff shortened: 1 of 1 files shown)


mambocab:
> flush failure can rebound w/raised exception

How do you mean 'rebound'?

Won't this break assumptions that, e.g., some tests make? I can see why successful flushing isn't strictly necessary for stopping a node in general, but it's part of `stop`'s behavior at this point.

josh-mckenzie:
"Rebound", as in raise an exception, bounce back out, and not stop the node in question.

Flush is no longer an assumption on .stop()'s behavior back since we switched it to taskkill on Windows due to errors shutting down conservatively. I'm shooting for a compromise here that essentially states "On Windows, we'll try to flush before we murder the process but no promises."

mambocab:
Gotcha, thanks for the explanation. Could you please update the comment to reflect that a successful `flush` isn't a hard requirement? Otherwise, I'm +1.

I don't have write access to this repo, so it may be a few days before it's merged.

josh-mckenzie:
Comment updated.

PR #362, opened 2015-08-20
Stress tool arguments not handled correctly

spodkowinski:
1. Stress tool options need to be _appended_ to argument list. Example for using stress tool directly:
`ccm stress user profile=...yaml 'ops(simple1=1)' -port jmx=7100`
1. jmx port will now be set automatically for node specific runs


ptnapoleon:
+1.

PR #364, opened 2015-08-20
Cluster readability

mambocab:
The most important commits here are the first and second. It was not fun at all to figure out what the arguments to `create_node` were, so those commits switch to using kwargs. I also made the selection of which port number to use for debugging use a plain old `foo if debug else bar` instead of some wacky `(foo, bar)[debug == True]` stuff.

While I was in there, I fixed up a bunch of linter errors.


ptnapoleon:
+1, thanks

PR #365, opened 2015-08-24
Removed kwargs from node.run_sstablesplit

pauloricardomg:
[PR363](https://github.com/pcmanus/ccm/pull/363) caused a [regression](http://cassci.datastax.com/view/cassandra-3.0/job/cassandra-3.0_dtest/lastCompletedBuild/testReport/sstablesplit_test/TestSSTableSplit/single_file_split_test/history/) on dtest `single_file_split_test` because that test specifies the `stdout` paramter on `kwargs`, what causes a conflict when calling `subprocess.Popen`, with `stdout=subprocess.PIPE`.

The fix is to provide and `output_file` parameter instead, similar to what is done in other tool methods, such as `run_sstablemetadata` or `run_sstableexpiredblockers`, so we can make a conditional on that before calling `subprocess.Popen`.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -818,7 +818,7 @@ class Node(object):
             subprocess.call(args, env=env)
 
     def run_sstablesplit(self, datafiles=None,  size=None, keyspace=None, column_families=None,
-                         no_snapshot=False, debug=False, **kwargs):
+                         no_snapshot=False, debug=False):
         sstablesplit = self._find_cmd('sstablesplit')
         env = common.make_cassandra_env(self.get_install_cassandra_root(), self.get_node_cassandra_root())
         sstablefiles = self.__gather_sstables(datafiles, keyspace, column_families)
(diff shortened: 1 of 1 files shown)


pauloricardomg:
The cassandra-dtest PR for using the new option on `single_file_split_test` is available [here](https://github.com/riptano/cassandra-dtest/pull/497).

mambocab:
@ptnapoleon has the final say on this, but I would much prefer this return a string than work directly with file handles. I know that that's how some of the existing methods work, but I don't like that behavior either :P

pauloricardomg:
@mambocab agreed.  `node.run_sstablesplit` already returns the standard output and error as string, so I modified the test to use that output instead of passing the `output_file` parameter.

Updated the PR to remove `kwargs`  from `run_sstablesplit`, since it is no longer necessary.

ptnapoleon:
+1

PR #366, opened 2015-08-24
Fix launch on Windows after CASSANDRA-9615

josh-mckenzie:
* Change windows launch to use node cassandra.ps1.
* Remove VerifyPortsAreAvailable on Windows startup
* Fix a flake8 violation in node.py

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -1009,7 +1009,7 @@ class Node(object):
             stress_options.append('-node')
             stress_options.append(self.address())
         # specify used jmx port if not already set
-        if not [ opt for opt in stress_options if opt.startswith('jmx=') ]:
+        if not [opt for opt in stress_options if opt.startswith('jmx=')]:
             stress_options.extend(['-port', 'jmx=' + self.jmx_port])
         args = [stress] + stress_options
         try:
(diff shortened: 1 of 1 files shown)


mambocab:
I haven't run this myself, but the code looksk good. If it works on your end, merge away.

ptnapoleon:
Sgtm, +1

ccmlib/node.py
```diff
@@ -1138,6 +1138,12 @@ def __clean_bat(self):
         common.replace_in_file(bat_file, run_pattern, "wmic process call create \"" + run_text + "\" > \"" +
                                self.get_path() + "/dirty_pid.tmp\"\n")
 
+        # On Windows, remove the VerifyPorts check from cassandra.ps1
+        common.replace_in_file(os.path.join(self.get_path(), 'bin', 'cassandra.ps1'), '        VerifyPortsAreAvailable', '')
```
jorgebay: I think this does not work for C* versions < 2.1

ptnapoleon: We don't run on Windows < 2.1 though?

jorgebay: We need ccm to support C* 1.2 and 2.0 for driver testing under Windows, like it used to.

PR #368, opened 2015-08-26
Lastlog

mambocab:
Adds a `showlastlog` cluster-level command to show `last.log`. I got annoyed when builds failed and I'd have to copy and paste the `last.log` path out of the error.

I also cleaned up `repository.py` a little to print the full `last.log` path. Now that I think about it, maybe those warnings should mention the new command.

In any event, I can iterate on this PR if you want the feature.


ptnapoleon:
I would take this as-is, but you're right, it would be nice if the warning mentioned the command.

mambocab:
Rebased improved error output into this.

PR #370, opened 2015-08-27
Added node.mark_log_for_errors() method

pauloricardomg:
The `mark_log_for_errors()` creates a mark in the current position of the cassandra log of the given node, where all error logs before that mark will be ignored in the `grep_log_for_errors()` method.

The main use case for this right now, is to ignore all errors on cassandra 2.1 on Windows before upgrading to 2.2, in the context of `upgrade_internal_auth_test.py` (CASSANDRA-10037).


ptnapoleon:
This is smart, I like it. +1

PR #371, opened 2015-09-02
Add jconsole command for cluster and nodes

spodkowinski:
Start jconsole and connect to running nodes using ccm.


ptnapoleon:
+1

PR #374, opened 2015-09-11
add versionfrombuild command

mambocab:
This gives a CLI command for getting the version of a node from its build.xml file. The name of the command isn't final; it looks kind of gross int the top-level usage statement:

```
$ ccm
Missing arguments
Usage:
  ccm <cluster_cmd> [options]
  ccm <node_name> <node_cmd> [options]

Where <cluster_cmd> is one of
  create         Create a new cluster
  add            Add a new node to the current cluster
[...]
or <node_name> is the name of a node of the current cluster and <node_cmd> is one of
[...]
  pause          Send a SIGSTOP to this node
  resume         Send a SIGCONT to this node
  jconsole       Opens jconsole client and connect to running node
  versionfrombuild Print the node's version as grepped from build.xml. Can be used when the node isn't running.
```

See that last line -- the command name spills over. Don't know if the name of the command should change, if the logic for displaying commands should change, or if we care at all.

But, this could reduce duplication -- we do this in bash in the dtests on CassCI.

### DIFF
--- a/ccmlib/cmds/node_cmds.py
+++ b/ccmlib/cmds/node_cmds.py
@@ -52,7 +52,8 @@ def node_cmds():
         "spark",
         "pause",
         "resume",
-        "jconsole"
+        "jconsole",
+        "versionfrombuild"
     ]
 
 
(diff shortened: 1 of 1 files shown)


ptnapoleon:
I don't really like the name, but it's fine. I would also like if, when the node is up, if we can compare the result of grepping the build.xml, with the output of nodetool version, to see if they match, and warn if not.

mambocab:
> I don't really like the name, but it's fine.

Samesies.

> I would also like if, when the node is up, if we can compare the result of grepping the build.xml, with the output of nodetool version, to see if they match, and warn if not.

Where should that logic live? I'd vote [Node.get_cassandra_version](https://github.com/pcmanus/ccm/blob/master/ccmlib/node.py#L215), maybe disabled by default and enabled with a kwarg.

ptnapoleon:
Just put it in the `run` method?

mambocab:
Yeah, that's simplest thing to do; thanks. Ready for another review.

ptnapoleon:
+1

PR #376, opened 2015-09-25
Modifications for DSE 4.8.0 support

bweck:
Added support for DSE 4.8.0.

* The dse jar filename changed from "dse-4.7.0.jar" to "dse-core-4.8.0.jar", modified get_dse_version to support new filename format.
* The dse 4.8.0 archive no longer contains ```resources/shark```, skip copying resources that don't exist.


ptnapoleon:
Thanks!

PR #379, opened 2015-10-01
Fixing spelling mistakes; this effects DECOMMISSIONED enumeration

mikefero:


ptnapoleon:
Thanks. I trust git blame won't hold me responsible for these, but better not to check :).

PR #381, opened 2015-10-05
Fix set_log_level after logging changes in CASSANDRA-10241

jkni:
In CASSANDRA-10241, we introduced a separate debug.log. As part of this, a logger for org.apache.cassandra was added in addition to the root level. In order to preserve the existing set_log_level semantics in CCM, we need to also set the log level for this org.apache.cassandra logger.


ptnapoleon:
Thanks!

PR #385, opened 2015-10-08
node.stress mutability fixes

knifewine:
don't use mutable default default arg
and don't do an in-place update to passed arg either

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -920,7 +920,12 @@ class Node(object):
                 files.remove(f)
         return files
 
-    def stress(self, stress_options=[], **kwargs):
+    def stress(self, stress_options=None, **kwargs):
+        if stress_options is None:
+            stress_options = []
+        else:
+            stress_options = stress_options[:]
+
         stress = common.get_stress_bin(self.get_install_dir())
         if self.cluster.cassandra_version() <= '2.1':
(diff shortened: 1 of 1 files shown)


ccmlib/node.py
```diff
@@ -920,7 +920,12 @@ def get_sstables(self, keyspace, column_family):
                 files.remove(f)
         return files
 
-    def stress(self, stress_options=[], **kwargs):
+    def stress(self, stress_options=None, **kwargs):
+        if stress_options is None:
+            stress_options = []
+        else:
+            stress_options = stress_options[:]
```
ptnapoleon: What is this for?

knifewine: To use a copy of a passed list rather than the list itself, since the method is doing in-place modification which can cause side effects in the calling code. I found this by accident when running node.stress multiple times with a passed in list (same list in the node.stress call multiple times) triggering syntax errors because -node was appended multiple times.

mambocab:
I'm +1. Another way to say `stress_options = stress_options[:]` is `stress_options = list(stress_options)`. They're equally idiomatic, so use whichever you think is clearer. (Using the `list` constructor is slightly slower, basically because it has to look up `list`, but that doesn't matter here.)

ptnapoleon:
+1

PR #389, opened 2015-10-09
Allowing for git branches to be used that resemble SHAs

mikefero:


ccmlib/repository.py
```diff
@@ -113,10 +113,32 @@ def clone_development(git_repo, version, verbose=False):
...
 
+                # determine if the request is for a branch
+                is_branch = False
+                try:
+                    branch_listing = subprocess.check_output(['git', 'branch', '--all'], cwd=target_dir)
+                    branch_listing = branch_listing.replace('remotes/origin/', '')
+                    for branch in branch_listing.splitlines(True):
+                        if git_branch == branch.strip():
+                            is_branch = True
+                except subprocess.CalledProcessError, cpe:
```
ptnapoleon: Is this also valid python 3 syntax?

ccmlib/repository.py
```diff
@@ -113,10 +113,32 @@ def clone_development(git_repo, version, verbose=False):
                 else:
                     subprocess.call(['git', 'clone', local_git_cache, target_dir], cwd=__get_dir(), stdout=lf, stderr=lf)
 
+                # determine if the request is for a branch
+                is_branch = False
+                try:
+                    branch_listing = subprocess.check_output(['git', 'branch', '--all'], cwd=target_dir)
+                    branch_listing = branch_listing.replace('remotes/origin/', '')
+                    for branch in branch_listing.splitlines(True):
+                        if git_branch == branch.strip():
```
ptnapoleon: It might be better to do `if git_branch in branch.strip():` here, just in case, no?

ccmlib/repository.py
```diff
@@ -113,10 +113,32 @@ def clone_development(git_repo, version, verbose=False):
...
+                            is_branch = True
```
ptnapoleon: Do we just want to break out of the for loop immediately if we find this?

ccmlib/repository.py
```diff
@@ -113,10 +113,32 @@ def clone_development(git_repo, version, verbose=False):
...
+                except subprocess.CalledProcessError, cpe:
+                        print_("Error Running Branch Filter: %s\nAssumming request is not for a branch" % cpe.output)
+
                 # now check out the right version
                 if verbose:
-                    print_("Checking out requested branch (%s)" % git_branch)
-                out = subprocess.call(['git', 'checkout', git_branch], cwd=target_dir, stdout=lf, stderr=lf)
+                    branch_or_sha_tag = 'SHA/tag'
+                    if is_branch:
+                        branch_or_sha_tag = 'branch'
```
ptnapoleon: You can condense these three lines down to:
`branch_or_sha_tag = 'branch' if is_branch else 'SHA/tag'`

mambocab:
> Is this also valid python 3 syntax?

It's not. There's also some additional finagling required for Python 3 compat; I can take care of that.

ccmlib/repository.py
mikefero: That would probably be better than the replace for sure.

ccmlib/repository.py
mikefero: short circuit indeed

ccmlib/repository.py
mikefero: +1; didn't see any use of the ternary operation when I did a quick glance so I opted not to do that

PR #390, opened 2015-10-13
allow sha like branch names

mambocab:
This adds some Python 3 compatability stuff on top of #389. It also addresses @ptnapoleon's comments on that PR.

@ptnapoleon can you review? and @mikefero I'd welcome your feedback if you want to have a look.

I tested this code on Python 2 and 3 using this script:

https://gist.github.com/mambocab/445012d260e71cfcbca6

### DIFF
--- a/ccmlib/repository.py
+++ b/ccmlib/repository.py
@@ -113,10 +113,32 @@ def clone_development(git_repo, version, verbose=False):
                 else:
                     subprocess.call(['git', 'clone', local_git_cache, target_dir], cwd=__get_dir(), stdout=lf, stderr=lf)
 
+                # determine if the request is for a branch
+                is_branch = False
+                try:
+                    branch_listing = subprocess.check_output(['git', 'branch', '--all'], cwd=target_dir)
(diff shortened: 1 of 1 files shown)


mikefero:
:+1: Thanks for taking on the review requests; much cleaner

ptnapoleon:
+1 on the code. Have you checked to make sure this works, even after the changes? I'm worried about substring matching with the `git_branch in branches` logic. Also, this helped with review, but Jim can you squash all your changes into one commit, so there are only 2 here?

mambocab:
> Have you checked to make sure this works, even after the changes?

Yep. Running the linked test script results in the output here (which I've cleaned up for readability; the full output is at the same gist):

https://gist.github.com/mambocab/99d9db4107ba3a0499d5#file-fero-out-shorter

`branches` is a list of strings, so `if branch in branches` checks membership, not substring matching.

mambocab:
And yes, happy to squash.

mambocab:
Squashed.

ptnapoleon:
Yes, I know it only checks membership, that was my fear. But if it works, then I'm merging.

PR #392, opened 2015-10-14
Windows fix

ptnapoleon:
@jorgebay and @josh-mckenzie to review

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -632,7 +632,12 @@ class Node(object):
                 break
             time.sleep(10)
 
-    def nodetool(self, cmd, capture_output=True):
+    def nodetool(self, cmd, capture_output=True, wait=True):
+        """
+        Setting wait=False makes it impossible to detect errors,
+        if capture_output is also False. wait=False allows us to return
+        while nodetool is still running.
+        """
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Only review the relevant commit. I accidentally included both. @mambocab, can you review the `wait=True` change?

josh-mckenzie:
+1 on VerifyPorts check

mambocab:
Before the `if wait:` block in this method, there's a

```python
        if capture_output:
            p = subprocess.Popen(args, env=env, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
            stdout, stderr = p.communicate()
```

I believe that `.communicate` blocks, so passing `capture_output=True` and `wait=False` would lead to unexpected behavior, where it waits on `communicate` even though `wait` was false. Maybe add an assertion that arguments don't contradict each other.

Other than that, +1.

jorgebay:
Tested on Windows for C* 1.2, 2.0, 3.0.0-rc1. ps1 / bat windows fix looks good!

Thanks @ptnapoleon

ptnapoleon:
@mambocab Is the new commit what youre looking for?

mambocab:
Yup, it is.

Issue #387, opened 2015-10-08
Determine desired set_log_level behavior 


jkni:
We recently introduced changes to C* that separate logging into a system.log that is filtered to INFO and above by default, as well as a debug.log that is filtered to DEBUG and above by default.

Before, set_log_level set the log level of a core root appender to the desired level. This would impact both C* and any other libraries and redirect this to system.log.

Now, set_log_level does not have a clearly defined behavior. I recently PRed a patch that sets the log level of this root appender and a newly added C* appender, but if set_log_level is called with a value higher than DEBUG (such as INFO), debug.log will only contain INFO entries (and not DEBUG) since the appender is set to INFO. This seems to defeat the point of a debug.log.

I propose the following changes, where we have an ascending scale of logging from ['trace', 'debug', 'info', 'warn', 'error'].

If set_log_level is called with 'trace', we set both appenders to 'trace'. Then, debug.log will contain all entries and system.log will contain only INFO and up.

If set_log_level is called with 'debug', we set both appenders to 'debug'. Then, debug.log will contain all entries debug and up and system.log will contain only INFO and up.

If set_log_level is called with 'info', this is a noop. Debug.log will contain C* entries debug and up and other libraries INFO and up, while system.log will contain INFO and up.

If set_log_level is called with 'warn', we set the system.log filter to WARN.

If set_log_level is called with 'error', we set the system.log filter to ERROR.

This is a decidedly more complicated behavior, but it seems to best fit the intention of the separate system.log and debug.log. Thoughts?

EDIT: I should note that I'm willing to implement this - I'd just like desired behavior to be decided upon first.

ptnapoleon:
I like the proposed behavior. I don't think any other users of ccm will be strongly affected by our changes, so we should mostly consider what is useful for testing. @mambocab, thoughts?

jkni:
Relevant PR merged.

PR #399, opened 2015-10-16
Add sane set_log_level behavior following CASSANDRA-10241

jkni:
This PR implements the set_log_level behavior proposed in Issue #387.

In short, recent changes to Cassandra 2.2.2 and above add a separate debug.log.  These broke CCM set_log_level. The previous fix simply set both log levels to the same level, which nullified much of the benefit of adding a separate debug.log. This fix adds behaviors that still retain a sensible amount of information in debug.log based on the log level set.

I've tested this with a few dtests and run it through autopep8.


ptnapoleon:
Seems like reasonable behavior to me, +1

PR #401, opened 2015-10-23
Retry version download from git in default case

ptnapoleon:
@mambocab to review.
Not finished testing yet, but what do you think?

### DIFF
--- a/ccmlib/repository.py
+++ b/ccmlib/repository.py
@@ -29,24 +29,47 @@ GITHUB_TAGS = "https://api.github.com/repos/apache/cassandra/git/refs/tags"
 
 def setup(version, verbose=False):
     binary = True
+    fallback = True
+
     if version.startswith('git:'):
         clone_development(GIT_REPO, version, verbose=verbose)
         return (version_directory(version), None)
+
     elif version.startswith('binary:'):
         version = version.replace('binary:', '')
+        fallback = False
+
     elif version.startswith('github:'):
         user_name, _ = github_username_and_branch_name(version)
         clone_development(github_repo_for_user(user_name), version, verbose=verbose)
         return (directory_name(version), None)
+
     elif version.startswith('source:'):
         version = version.replace('source:', '')
         binary = False
+        fallback = False
+
     if version in ('stable', 'oldstable', 'testing'):
         version = get_tagged_version_numbers(version)[0]
+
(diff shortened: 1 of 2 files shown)


ccmlib/repository.py
```diff
@@ -29,24 +29,46 @@
...
         version = get_tagged_version_numbers(version)[0]
+
     cdir = version_directory(version)
     if cdir is None:
-        download_version(version, verbose=verbose, binary=binary)
-        cdir = version_directory(version)
+        try:
+            download_version(version, verbose=verbose, binary=binary)
+            cdir = version_directory(version)
+        except Exception as e:
```
mambocab: I think I'd move this retry/fallback logic to download_version, e.g. `download_version(version, verbose=verbose, binary=binary, fallback=fallback)`. That's just my gut talking, though, I haven't got a super-strong argument for it.

ptnapoleon: if clone_development weren't already a distinct function, I would agree with you. But since the pseudocode here is:
```
try download_version
if that fails, run clone_development
```
It feels better to keep the logic out here.

mambocab: Ah, I see, I believe I'm ok with that.

ccmlib/repository.py
```diff
@@ -29,24 +29,46 @@
...
+            # If we failed to download from ARCHIVE,
+            # then we build from source from the git repo,
+            # as it is more reliable.
+            # We don't do this if binary: or source: were
+            # explicitly specified.
+            if fallback:
+                version = 'git:cassandra-{}'.format(version)
```
mambocab: Are we assuming that if the binary download doesn't work, that the source download won't either? I would think the most complete fallback path would be more like archive binary -> archive source -> Apache Git. (Maybe even Apache GitHub after that.) But this looks like it falls straight from binary to Apache Git; is there anything wrong with trying source for even greater redundancy?

ptnapoleon: Other than code complexity? Not really, but it does seem that if the archive lacks the binary for version X, it will lack the source. The files are uploaded side by side at the same time.

mambocab:
nice, +1

ptnapoleon:
Awesome. I'm going to merge in the morning, so as not to cause any late friday night breakages in the test suites that use this.

PR #403, opened 2015-10-30
py3 fixes

nutbunnies:


ccmlib/node.py
```diff
@@ -547,8 +547,8 @@ def start(self,
 
         if verbose:
             stdout, stderr = process.communicate()
-            print stdout
-            print stderr
+            print(stdout)
+            print(stderr)
```
ptnapoleon: These should both be `print_()` not `print()`

PR #404, opened 2015-10-30
Add hints_dir to conf based on node, not cluster version

beobal:
During rolling upgrades, the cluster version remains on the old
version as individual nodes are updated. When a node is upgraded
to 3.0+, this leads to the hints_directory setting not being added
to its config. In turn, this means that the hints dir defaults to
a location under CASSANDRA_HOME which is shared by all nodes in
he cluster. From that, all manner of hijinks ensue with nodes
wiping out each other's hints files, causing seemingly random
exceptions and other fun stuff.


ptnapoleon:
Seems like a good change, thanks

PR #406, opened 2015-11-10
Only update node configuration when necessary

pauloricardomg:
Currently every call to `common.make_cassandra_env()` updates the node cassandra-env file with the current install dir and node conf dir. This method is used by many `node` commands, such as `nodetool`, `json2sstable`, `scrub`, `cqlsh`, etc to construct the environment variables, but it's not necessary to update the cassandra-env file every time these commands are called, only in the first time, or when the install dir changes.

This patch makes all node commands use `node.get_env()` instead of  `common.make_cassandra_env()`, and only updates the cassandra-env file in the first use, or when the install dir changes.

The motivation behind this is a race condition in the context of [CASSANDRA-10485](https://issues.apache.org/jira/browse/CASSANDRA-10485), where a nodetool command is called concurrently by 2 threads, so both threads try to modify the cassandra-env file concurrently (while it was not necessary in the first place).

It would be nice to trigger a dtest run with this patch to see if there are any unintended side-effects, but I did a quick look and it seems fine.

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -175,7 +175,7 @@ def rmdirs(path):
         shutil.rmtree(path)
 
 
-def make_cassandra_env(install_dir, node_path):
+def make_cassandra_env(install_dir, node_path, update_conf=True):
     if is_win() and get_version_from_build(node_path=node_path) >= '2.1':
         sh_file = os.path.join(CASSANDRA_CONF_DIR, CASSANDRA_WIN_ENV)
     else:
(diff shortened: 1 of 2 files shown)


mambocab:
I don't know enough about this code to review this quickly (@ptnapoleon?) but: I appreciate all the information in the PR, could you add it to the commit message as well?

pauloricardomg:
I don't know enough either, heh. Added the full description to the commit. :+1:

ptnapoleon:
Well it all seems good. I'll set up a dtest run, as I don't want to break anything. Thanks

pauloricardomg:
Thanks @ptnapoleon

ptnapoleon:
Test results appear the same as default C*. Merging. Thanks @pauloricardomg !

PR #407, opened 2015-11-10
PEP8 Fixes

lmr:
I was taking a look at ccm's code with a checking tool I wrote (similar to pyflakes) that helped me prepare some PEP8, pylint and other fixes. If you guys think it's interesting, I could also send a PR to perform an automated CI job on each PR through Travis CI (this way, the amount of lint/pep8 errors that get into CCM master will reduce substantially).

### DIFF
--- a/ccm
+++ b/ccm
@@ -1,11 +1,13 @@
 #!/usr/bin/env python
 
-import os, sys
+import os
+import sys
 
 from six import print_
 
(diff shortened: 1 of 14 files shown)


ptnapoleon:
Thanks for the cleanup! We're aware of what needs done for Travis, but just haven't gotten around to setting it up yet.

PR #408, opened 2015-11-11
Use quiet flag for windows

jorgebay:
[Cassandra 3.0 on Windows uses `-NoNewWindow` argument by default](https://github.com/apache/cassandra/commit/f8783b7bb1bd16d29565c3be86480a7188e91ee0#diff-67909d13f7884e965e96d1df987e43ebR263). When starting ccm from another process, it causes `ccm start` to never return as cassandra subprocess is running on the same window.

Using cassandra.ps1 -q flag solves the issue.

cc @josh-mckenzie

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -527,6 +527,8 @@ class Node(object):
         if use_jna is False:
             args.append('-Dcassandra.boot_without_jna=true')
         env['JVM_EXTRA_OPTS'] = env.get('JVM_EXTRA_OPTS', "") + " " + " ".join(jvm_args)
+        if common.is_win():
+            args.append('-q')
 
         # In case we are restarting a node
         # we risk reading the old cassandra.pid file


josh-mckenzie:
-1. We changed this specifically to fix https://issues.apache.org/jira/browse/CASSANDRA-10641, so if we revert that in ccm it's going to re-break those dtests. We need to take a -q flag on ccm and pass it through on Windows, then push the responsibility for using that upstream to apps that expect that behavior.

jorgebay:
Having a -q flag on `ccm start` sounds good to me.

ptnapoleon:
@josh-mckenzie Is it only those dtests that need the loud behavior? Could we not just default to `-q` unless specified otherwise on start?

josh-mckenzie:
I think the real problem is (and this is totally my bad) - having the quiet behavior by default is a deviation from the behavior on linux, and it adds an extra point-of-failure for multi-platform test writing. If we just said "For all tests, grep the logs for things, don't rely on stdout/stderr" it'd also fix this mismatch, but the reality is that the more in parity we can have these 2 platforms the better.

I'm totally good with adding a -q flag on ccm start that just does a simple pass-through, and again - my bad on creating this mismatch in the first place. :)

ptnapoleon:
@jorgebay Do you mind if I close this in favor of a different PR?

jorgebay:
Closing

PR #409, opened 2015-11-11
Add option for quiet start on windows

ptnapoleon:
@josh-mckenzie and @mambocab to review.

@jorgebay This is what I opened in place of #408.

### DIFF
--- a/ccmlib/cmds/cluster_cmds.py
+++ b/ccmlib/cmds/cluster_cmds.py
@@ -521,7 +521,7 @@ class ClusterStartCmd(Cmd):
                           help="Start the nodes with yourkit agent (only valid with -s)", default=False)
         parser.add_option('--profile-opts', type="string", action="store", dest="profile_options",
                           help="Yourkit options when profiling", default=None)
-        parser.add_option('--quiet-windows', action="store_true", dest="quiet_start", help="Pass -q on Windows 2.2.4+ startup. Ignored on linux.", default=False)
+        parser.add_option('--quiet-windows', action="store_true", dest="quiet_start", help="Pass -q on Windows 2.2.4+ and 3.0+ startup. Ignored on linux.", default=False)
         return parser
 
     def validate(self, parser, options, args):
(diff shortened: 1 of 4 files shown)


ptnapoleon:
Josh, do I need to version gate this? I'll start testing this on windows now

josh-mckenzie:
Gate as per 10641, or 2.2.4/3.0.0 or higher

ptnapoleon:
Okay, ready for review, but in all honestly, I don't see any difference in behavior from starting a three node cluster on Windows.

mambocab:
Code looks fine, though I haven't run it. 

Do we want to fail or print a warning if you pass `--quiet-windows` on Linux? I know it's got `windows` in the name, but still, being conservative about accepting arguments is worth considering.

ptnapoleon:
Right now we just drop it? I'd rather warn at worst, as this way we don't have to branch on which OS we pass this in on, for the tests that need it

ptnapoleon:
Lmk what you think of the latest commit

ccmlib/cmds/cluster_cmds.py
```diff
@@ -521,6 +521,7 @@ def get_parser(self):
                           help="Start the nodes with yourkit agent (only valid with -s)", default=False)
         parser.add_option('--profile-opts', type="string", action="store", dest="profile_options",
                           help="Yourkit options when profiling", default=None)
+        parser.add_option('--quiet-windows', action="store_true", dest="quiet_start", help="Pass -q on Windows 2.2.4+ startup. Ignored on linux.", default=False)
```
josh-mckenzie: 2.2.4+ and 3.0+?

ccmlib/cmds/node_cmds.py
```diff
@@ -183,6 +183,7 @@ def get_parser(self):
                           help="Replace a node in the ring through the cassandra.replace_address option")
         parser.add_option('--jvm_arg', action="append", dest="jvm_args",
                           help="Specify a JVM argument", default=[])
+        parser.add_option('--quiet-windows', action="store_true", dest="quiet_start", help="Pass -q on Windows 2.2.4+  startup. Ignored on linux.", default=False)
```
josh-mckenzie: Same as above - might be worth mentioning 3.0+

josh-mckenzie:
other than the 2 nits, lgtm

mambocab:
The WARN looks good to me. :+1:

PR #410, opened 2015-11-12
add watch_log_for_listening method

mambocab:
Adds a method, like watch_log_for_alive, that waits for the node it's called on to be listening for CQL clients. `wait_for_binary_interface` isn't what I want because it also times out checking for a listening socket, which is not what I want to do.

(Incidentally, I don't understand the design of `wait_for_binary_interface`. If the socket isn't listening, why print a warning rather than raise an exception?)

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -440,6 +440,15 @@ class Node(object):
         tofind = ["%s.* now UP" % node.address() for node in tofind]
         self.watch_log_for(tofind, from_mark=from_mark, timeout=timeout, filename=filename)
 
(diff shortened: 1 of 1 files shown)


ptnapoleon:
I don't see the benefit of this over `wait_for_binary_interface`. Also, I don't know why we warn there.

mambocab:
The benefit is that you can actually set the timeout for the whole method, rather than also having the additional timeout listening for the socket.

ptnapoleon:
I'd rather modify wait_for_binary_interface?

ptnapoleon:
Did we ultimately decide we didn't need this?

PR #411, opened 2015-11-17
Check for a bad java.exe that is causing some user issues

ptnapoleon:
@josh-mckenzie Is this what you had in mind?


josh-mckenzie:
+1

PR #415, opened 2015-11-24
DSE Version Detection broken < DSE 4.8.0

MightyE:
Fix for https://github.com/pcmanus/ccm/pull/376/files#diff-e19f0f0ec871623be418f8c4e97f9f46L530

The updated regex in that commit only supports DSE >= 4.8.0 (where the jar file name changed to dse-core-X.Y.Z.jar), and fails for earlier DSE versions (where the jar file name was dse-X.Y.Z.jar).


ptnapoleon:
Thanks!

PR #417, opened 2015-11-30
ccmlib: Avoid passing mutable default argument to jvm_args

lmr:
Change jvm_args default value to None, then in the method
body check if it's None, then assign it to an empty list.
This avoids surprises if multiple calls to the methods
are made [1].

[1] http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments


ptnapoleon:
+1, thanks!

PR #418, opened 2015-11-30
Remove all usages of empty lists in kwarg defaults

ptnapoleon:
@mambocab to review. #417 brought to my attention that this mistake was occurring in our codebase.


mambocab:
Looks good - the only instances I see of changing `[]` to `None` and not subsequently setting it are when the method isn't implemented. +1.

PR #420, opened 2015-12-02
Adding command line arg for credentials file.

GregBestland:
This change would allow for users to provide a credentials file that contains dse_username and dse_password to use for credentials instead of passing them in explicitly.

I think this change would be useful for anyone who is trying to share common credentials, or keep credentials out of logs. 

The format of the credentials file would be as follows.

[dse_credentials]
dse_username = username
dse_password = password


ptnapoleon:
Lgtm, Thanks Greg!

PR #426, opened 2015-12-16
Use ~/.ccm/.dse.ini as dse credentials file if it exists by default.

tolbertam:
I really like the recent addition of the '--dse-credentials' config parameter for loading credentials from a config file.   For convenience I think it would be nice if there were a default credentials file that, if exists, will be loaded to grab the dse credentials.   This would remove the need to pass the path of the credentials file each time you create a cluster.

The existing behavior:

* If user passes in --dse-credentials, grab dse_username and dse_password from the passed in config file.
* Else grab from --dse-username and/or --dse-password

The new behavior:

* If user doesn't pass in --dse-credentials and .dse.ini exists, set dse_username and dse_password based on it's contents.
* Else If user passes in --dse-credentials, grab dse_username and dse_password from the passed in config file.
* If --dse-username and/or --dse-password were passed in explicitly, override the dse_username and dse_password values.

### DIFF
--- a/ccmlib/cmds/cluster_cmds.py
+++ b/ccmlib/cmds/cluster_cmds.py
@@ -79,7 +79,7 @@ class ClusterCreateCmd(Cmd):
         parser.add_option("--dse-password", type="string", dest="dse_password",
                           help="The password to use to download DSE with", default=None)
         parser.add_option("--dse-credentials", type="string", dest="dse_credentials_file",
-                          help="A file containing the dse_username, and dse_password.", default=None)
+                          help="An ini-style config file containing the dse_username and dse_password under a dse_credentials section. [default to {0}/.dse.ini if it exists]".format(common.get_default_path_display_name()), default=None)
         parser.add_option("--install-dir", type="string", dest="install_dir",
                           help="Path to the cassandra or dse directory to use [default %default]", default="./")
         parser.add_option('-n', '--nodes', type="string", dest="nodes",
--- a/ccmlib/cmds/cluster_cmds.py
+++ b/ccmlib/cmds/cluster_cmds.py
@@ -79,7 +79,7 @@ class ClusterCreateCmd(Cmd):
         parser.add_option("--dse-password", type="string", dest="dse_password",
                           help="The password to use to download DSE with", default=None)
         parser.add_option("--dse-credentials", type="string", dest="dse_credentials_file",
-                          help="An ini-style config file containing the dse_username and dse_password under a dse_credentials section. [default to {0}/.dse.ini if it exists]".format(common.get_default_path_display_name()), default=None)
+                          help="An ini-style config file containing the dse_username and dse_password under a dse_credentials section. [default to {}/.dse.ini if it exists]".format(common.get_default_path_display_name()), default=None)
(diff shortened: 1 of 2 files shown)


ccmlib/dse_cluster.py
```diff
@@ -18,10 +18,12 @@
...
     def __init__(self, path, name, partitioner=None, install_dir=None, create_directory=True, version=None, dse_username=None, dse_password=None, dse_credentials_file=None, opscenter=None, verbose=False):
-        if dse_credentials_file:
-            self.load_credentials_from_file(dse_credentials_file)
-        else:
+        self.dse_username = None
+        self.dse_password = None
+        self.load_credentials_from_file(dse_credentials_file)
+        if dse_username is not None:
             self.dse_username = dse_username
+        if dse_password is not None:
```
ptnapoleon: Is there a need to define self.dse_username and password as None? Couldn't this just be
```
self.load_credentials_from_file(dse_credentials_file)
if dse_username is not None:
    self.dse_username, self.dse_password = dse_username, dse_password 
```

ccmlib/cmds/cluster_cmds.py
```diff
@@ -79,7 +79,7 @@ def get_parser(self):
         parser.add_option("--dse-password", type="string", dest="dse_password",
                           help="The password to use to download DSE with", default=None)
         parser.add_option("--dse-credentials", type="string", dest="dse_credentials_file",
-                          help="A file containing the dse_username, and dse_password.", default=None)
+                          help="An ini-style config file containing the dse_username and dse_password under a dse_credentials section. [default to {0}/.dse.ini if it exists]".format(common.get_default_path_display_name()), default=None)
```
ptnapoleon: I am fairly sure you can just use `{}` instead of `{0}` in your format string.

ccmlib/dse_cluster.py
tolbertam: There would be a possible issue if there is no credentials file and --dse-username and --dse-password were not passed in, here is the failure behavior with explicitly setting them to None beforehand:

```
ccmlib.common.ArgumentError: Invalid url http://downloads.datastax.com/enterprise/dse-4.8.2-bin.tar.gz (underlying error is: HTTP Error 401: Unauthorized)
```

And without:

```
AttributeError: 'DseCluster' object has no attribute 'dse_username'
```

I think the first scenario is more obvious, but maybe we should assert that there was a resolved username / password beforehand.  I'll add some additional logic for that.

tolbertam: I've improved this by logging warnings if either dse_username or password are not set:

```
Warning: No dse username detected, specify one using --dse-username or passing in a credentials file using --dse-credentials.
Warning: No dse password detected, specify one using --dse-password or passing in a credentials file using --dse-credentials.
```

The download will still be attempted (and thus the Invalid url error will appear), I figured i'd leave it as is in the event the download ever doesn't require authentication.

ccmlib/cmds/cluster_cmds.py
tolbertam: Thanks, fixed :)

ptnapoleon:
+1

Issue #419, opened 2015-12-01
Appears we cannot run DSE 4.7+ anymore


ptnapoleon:
This should be investigated, and hopefully fixed.

marshallk:
You can run DSE 4.7+ .. but a lot of commands and logs do not work well. For example:

1. ccm start - This command will not return..It will start up the nodes after I ctrl+c out. If I just run the ccm start and do nothing.. it wll not return. I've noticed that it will start the first node and not the rest until I ctrl+c out. I believe it is in a loop with a timeout..dealing with log files.

2. ccm <node> stop will only work if I start the nodes one by one with the ccm node start.

In general the logging commands don't work.

MightyE:
Possibly related to apparently system.log never getting created when started with `ccm start`?  Don't you watch system.log for a success message to know when to return?

ptnapoleon:
Yes, that is precisely the issue at hand, or at least the primary one.

kishkaru:
Tested DSE 4.0 through 4.8 and all work fine with the patch in #427.

PR #427, opened 2015-12-17
Appears we cannot run DSE 4.7+ anymore #419

kais92:
The -Dcassandra.logdir= is set to  "/var/log/cassandra" by default 
This patch set the right path. 
It's work with the last dse (4.8.3)


ptnapoleon:
What effect does this have on versions prior to 4.7?

kais92:
Hi,
It works for dse 4.6.9

ptnapoleon:
Sounds good to me. Thanks for the patch

PR #428, opened 2015-12-18
Only log dse credentials are missing when attempting to download.

tolbertam:
It looks like I introduced an annoying regression in #426.   The log messages I added for informing a user that they were not providing DSE credentials were evaluated when running *any* command against a DSE cluster.   This should really only be logged before you go to download a DSE version.   This makes that adjustment so the log message only appears when it is relevant.   

Props to @kishkaru for noticing this :).


ptnapoleon:
+1!

PR #431, opened 2015-12-22
Allow passing in literal yaml to updateconf

ptnapoleon:
@tolbertam to review


ptnapoleon:
```
a:
- b:
  - c
  - d
```

achieved by
`ccm updateconf -y 'a: [b: [c,d]]'`

tolbertam:
Sorry I haven't had a chance to look at this recently, will take a look soon.

thibaultCha:
This would be a very handy feature :+1:

ptnapoleon:
@mambocab Can you review when you get a chance?

mambocab:
I've confirmed that you get the expected behavior with `ccm updateconf -y 'a: [b: [c,d]]'` and `ccm updateconf -y 'auto_bootstrap: true'`. Code looks good too. You might reword the documentation a bit -- something like "If enabled, treat argument as literal yaml, not key-value pairs". It's not obvious that `-y` is a flag that flips a bit, rather than an option that itself takes an argument.

PR #432, opened 2015-12-28
Avoid potential deadlock when running cqlsh

stef1927:
As described in the [documentation] (https://docs.python.org/2.7/library/subprocess.html#subprocess.Popen.wait), using `p.wait()` may result in a deadlock. This seems to happen quite regularly on Windows, although I did see it once on Linux too. The solution is quite simple, just use `p.communicate()` instead. This should be safe provided the input and output can be buffered in memory.


ptnapoleon:
Thanks Stefania.

PR #437, opened 2016-01-07
Start fixes

ptnapoleon:
@mambocab and @kishkaru to review. I'm running dtest against this right now to make sure I didn't break anything. It might be worthwhile to run a driver CI build against this? If that's something easily do-able.

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -1,5 +1,6 @@
 # ccm clusters
 
+import itertools
 import os
 import random
 import shutil
@@ -269,14 +270,11 @@ class Cluster(object):
             else:
                 node.show(only_status=True)
 
-    def start(self, no_wait=False, verbose=False, wait_for_binary_proto=False, wait_other_notice=False, jvm_args=None, profile_options=None, quiet_start=False, allow_root=False):
+    def start(self, no_wait=False, verbose=False, wait_for_binary_proto=False, wait_other_notice=True, jvm_args=None, profile_options=None, quiet_start=False, allow_root=False):
         if jvm_args is None:
             jvm_args = []
 
         common.assert_jdk_valid_for_cassandra_version(self.cassandra_version())
-
-        if wait_other_notice:
-            marks = [(node, node.mark_log()) for node in list(self.nodes.values())]
 
         started = []
         for node in list(self.nodes.values()):
@@ -288,7 +286,7 @@ class Cluster(object):
                 p = node.start(update_pid=False, jvm_args=jvm_args, profile_options=profile_options, verbose=verbose, quiet_start=quiet_start, allow_root=allow_root)
                 started.append((node, p, mark))
 
-        if no_wait and not verbose:
+        if no_wait:
             time.sleep(2)  # waiting 2 seconds to check for early errors and for the pid to be set
         else:
             for node, p, mark in started:
@@ -304,23 +302,14 @@ class Cluster(object):
             if not node.is_running():
                 raise NodeError("Error starting {0}.".format(node.name), p)
 
-        if not no_wait and self.cassandra_version() >= "0.8":
-            # 0.7 gossip messages seems less predictible that from 0.8 onwards and
-            # I don't care enough
-            for node, _, mark in started:
-                for other_node, _, _ in started:
-                    if other_node is not node:
-                        node.watch_log_for_alive(other_node, from_mark=mark)
+        if not no_wait:
+            if wait_other_notice:
+                for (node, _, mark), (other_node, _, _) in itertools.permutations(started, 2):
+                    node.watch_log_for_alive(other_node, from_mark=mark)
 
-        if wait_other_notice:
-            for old_node, mark in marks:
-                for node, _, _ in started:
-                    if old_node is not node:
-                        old_node.watch_log_for_alive(node, from_mark=mark)
-
-        if wait_for_binary_proto:
-            for node, p, mark in started:
-                node.wait_for_binary_interface(process=p, verbose=verbose, from_mark=mark)
+            if wait_for_binary_proto:
+                for node, p, mark in started:
+                    node.wait_for_binary_interface(process=p, verbose=verbose, from_mark=mark)
 
         return started
 
--- a/ccmlib/cmds/cluster_cmds.py
+++ b/ccmlib/cmds/cluster_cmds.py
@@ -524,9 +524,12 @@ class ClusterStartCmd(Cmd):
         parser.add_option('-v', '--verbose', action="store_true", dest="verbose",
                           help="Print standard output of cassandra process", default=False)
         parser.add_option('--no-wait', action="store_true", dest="no_wait",
-                          help="Do not wait for cassandra node to be ready", default=False)
-        parser.add_option('--wait-other-notice', action="store_true", dest="wait_other_notice",
-                          help="Wait until all other live nodes of the cluster have marked this node UP", default=False)
+                          help="Do not wait for cassandra node to be ready. Overrides all other wait options.", default=False)
+        # This option (wait-other-notice) is now deprecated, as it was never respected
+        parser.add_option('--wait-other-notice', action="store_true", dest="deprecate",
+                          help="DEPRECATED/IGNORED: Use '--skip-wait-other-notice' instead. This is now on by default.", default=False)
+        parser.add_option('--skip-wait-other-notice', action="store_false", dest="wait_other_notice",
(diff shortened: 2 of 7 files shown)


ptnapoleon:
Cassci alleges this breaks nothing, which is nice

kishkaru:
Didn't fix it for me. Now it hangs indefinitely at start.
Try creating a DSE cluster, start it, stop and re-start it. It hangs when it attempts to re-start (node1 doesn't complete starting). As soon as I kill the CCM python process, node1 starts right up. If it doesn't repro, try stopping and starting a few times.

Tested with DSE 4.7.6 and 4.8.3

ptnapoleon:
What are the commands you are using? I'm able to create 4.8.3, and start/stop indefinitely without a problem.

ptnapoleon:
@mambocab Can you review? This doesn't fix Kishan's issue, but it still a fix I would like to merge. Cassci isn't upset about it.

ccmlib/cluster.py
```diff
@@ -304,23 +301,19 @@ def start(self, no_wait=False, verbose=False, wait_for_binary_proto=False, wait_
...
-        if wait_for_binary_proto:
-            for node, p, mark in started:
-                node.wait_for_binary_interface(process=p, verbose=verbose, from_mark=mark)
+        if not no_wait:
+            if wait_other_notice:
+                # For each node
+                for node, _, mark in started:
+                    # Go through every other node
+                    for other_node, _, _ in started:
+                        if other_node is not node:
```
mambocab: [`itertools.permutations`](https://docs.python.org/2/library/itertools.html#itertools.permutations) will handle this nested-loop-and-filter logic for you:

```python
>>> import itertools, pprint
>>> pprint.pprint(list(itertools.permutations('ABCD', 2)))
[('A', 'B'),
 ('A', 'C'),
 ('A', 'D'),
 ('B', 'A'),
 ('B', 'C'),
 ('B', 'D'),
 ('C', 'A'),
 ('C', 'B'),
 ('C', 'D'),
 ('D', 'A'),
 ('D', 'B'),
 ('D', 'C')]
```

ccmlib/cmds/cluster_cmds.py
```diff
@@ -524,9 +524,12 @@ def get_parser(self):
         parser.add_option('-v', '--verbose', action="store_true", dest="verbose",
                           help="Print standard output of cassandra process", default=False)
         parser.add_option('--no-wait', action="store_true", dest="no_wait",
-                          help="Do not wait for cassandra node to be ready", default=False)
-        parser.add_option('--wait-other-notice', action="store_true", dest="wait_other_notice",
-                          help="Wait until all other live nodes of the cluster have marked this node UP", default=False)
+                          help="Do not wait for cassandra node to be ready. Overrides all other wait options.", default=False)
```
mambocab: I'm a little worried about silent collisions between the wait options -- should the CLI raise an error if you use `--wait-for-binary-proto` and `--no-wait`?

ccmlib/cmds/cluster_cmds.py
```diff
@@ -524,9 +524,12 @@ def get_parser(self):
...
+        # This option (wait-other-notice) is now deprecated, as it was never respected
+        parser.add_option('--wait-other-notice', action="store_true", dest="deprecate",
+                          help="DEPRECATED/IGNORED: Use '--skip-wait-other-notice' instead.", default=True)
```
mambocab: This `help` doesn't mention that it's on by default now.

ccmlib/cmds/cluster_cmds.py
mambocab: To be clear, by "silent collisions", I mean situations where someone uses both and is surprised by the effects of `--wait-for-binary-proto` not working.

ptnapoleon:
Okay, ready for re-review, @mambocab

mambocab:
Haven't looked at the changes again, but: looks like everything got squashed into a single commit. I liked the commits as they were; can you change it back?

ccmlib/cluster.py
```diff
@@ -304,23 +302,14 @@ def start(self, no_wait=False, verbose=False, wait_for_binary_proto=False, wait_
...
-                    if old_node is not node:
-                        old_node.watch_log_for_alive(node, from_mark=mark)
-
-        if wait_for_binary_proto:
-            for node, p, mark in started:
-                node.wait_for_binary_interface(process=p, verbose=verbose, from_mark=mark)
+        if not no_wait:
+            if wait_other_notice:
+                for (node, _, mark), (other_node, _, _) in itertools.permutations(started, 2):
+                            node.watch_log_for_alive(other_node, from_mark=mark)
```
mambocab: Indentation on this line is too deep.

kishkaru:
Looks good to me. It didn't fix the issues with DSE starting, but it did clean up the code to what it actually needs to do.

ccmlib/cmds/cluster_cmds.py
```diff
@@ -541,6 +544,8 @@ def get_parser(self):
 
     def validate(self, parser, options, args):
         Cmd.validate(self, parser, options, args, load_cluster=True)
+        if self.options.no_wait and self.wait_for_binary_proto:
+            print_("WARN: Both --wait-for-binary-proto and --no-wait were specified. --no-wait will be preferred.")
```
mambocab: Nice. Maybe do the same for the deprecated `--wait-other-notice` option?

mambocab:
Re-review done. Address my 2 comments, and if you can, split this back out into a pep8 commit, a DSE timeout bump commit, and a wait overhaul commit, and I'm +1.

ccmlib/cmds/cluster_cmds.py
ptnapoleon: I consciously chose not to do that since --wait-other-notice is true by default. Meaning you would get that warning everytime you selected `--no-wait` unless you ALSO did `--skip-wait-other-notice`

mambocab: Then I'd say the thing to do is have the behavior on by default, but have the flag's default be `False`, so you can detect when the flag is passed, yes?

ptnapoleon: That is what happened. Except I had to leave the --wait-other-notice for backwards compatibility.

ccmlib/cmds/cluster_cmds.py
```diff
@@ -524,9 +524,12 @@ def get_parser(self):
...
                           help="Print standard output of cassandra process", default=False)
         parser.add_option('--no-wait', action="store_true", dest="no_wait",
-                          help="Do not wait for cassandra node to be ready", default=False)
-        parser.add_option('--wait-other-notice', action="store_true", dest="wait_other_notice",
-                          help="Wait until all other live nodes of the cluster have marked this node UP", default=False)
+                          help="Do not wait for cassandra node to be ready. Overrides all other wait options.", default=False)
+        # This option (wait-other-notice) is now deprecated, as it was never respected
+        parser.add_option('--wait-other-notice', action="store_true", dest="deprecate",
+                          help="DEPRECATED/IGNORED: Use '--skip-wait-other-notice' instead. This is now on by default.", default=True)
+        parser.add_option('--skip-wait-other-notice', action="store_false", dest="wait_other_notice",
```
mambocab: Just noticed -- naming the destination of this flag `wait_other_notice` is very confusing.

ccmlib/cmds/cluster_cmds.py
mambocab: Sorry, I mean this:

https://github.com/mambocab/ccm/commit/0aa0891fc3d163b220ec36514cb2ee3b39dc3de2

or... something like that. I, um, haven't actually run it, and I don't remember how argparse works right now.

ptnapoleon:
Alright, all complaints resolved, probably. Ready for another look from @mambocab

ccmlib/cmds/cluster_cmds.py
```diff
@@ -541,6 +544,11 @@ def get_parser(self):
 
     def validate(self, parser, options, args):
         Cmd.validate(self, parser, options, args, load_cluster=True)
+        if self.options.deprecate:
+            print_("WARN: --wait-other-notice is deprecated. Please see the help text.")
+        if self.options.no_wait and (self.options.wait_for_binary_proto or self.options.deprecate):
+            print_("ERROR: --no-wait was specified alongside one or more wait options. This is invalid.")
```
mambocab: Beautiful.

mambocab:
Looks good all my comments were addressed. +1, thanks!

PR #438, opened 2016-01-07
Set gc.log location properly per node

ptnapoleon:
@mambocab to review

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -1427,6 +1427,17 @@ class Node(object):
         if self.get_cassandra_version() < '2.0.1':
             common.replace_in_file(conf_file, "-Xss", '    JVM_OPTS="$JVM_OPTS -Xss228k"')
 
+        # gc.log was turned on by default in 2.2.5/3.0.3/3.3
+        if self.get_cassandra_version() >= '2.2.5':
+            gc_log_pattern = "-Xloggc"
+            gc_log_path = os.path.join(self.get_path(), 'logs', 'gc.log')
+            if common.is_win():
+                gc_log_setting = '    $env:JVM_OPTS="$env:JVM_OPTS -Xloggc:{}"'.format(gc_log_path)
+            else:
+                gc_log_setting = 'JVM_OPTS="$JVM_OPTS -Xloggc:{}"'.format(gc_log_path)
+
+            common.replace_in_file(conf_file, gc_log_pattern, gc_log_setting)
+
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Despite 10140 only being in 2.2.5/3.0.3/3.3, and me checking for 2.2.5+, this does not break 3.0.0-3.0.1, or 3.1-3.2.

ptnapoleon:
Cassci alleges I broke nothing

ccmlib/node.py
```diff
@@ -1427,6 +1427,16 @@ def __update_envfile(self):
         if self.get_cassandra_version() < '2.0.1':
             common.replace_in_file(conf_file, "-Xss", '    JVM_OPTS="$JVM_OPTS -Xss228k"')
 
+        # gc.log was turned on by default in 2.2.5/3.0.3/3.3
+        if self.get_cassandra_version() >= '2.2.5':
+            gc_log_pattern = "-Xloggc"
+            if common.is_win():
+                gc_log_setting = '    $env:JVM_OPTS="$env:JVM_OPTS -Xloggc:$env:{}/logs/gc.log"'.format(self.get_path())
```
mambocab: What's with the leading spaces in this string?

mambocab:
Have you run this on Windows as well?

ccmlib/node.py
ptnapoleon: There's indentation in the file that I need to respect.

ptnapoleon:
I will right now

ptnapoleon:
Works on Windows

ccmlib/node.py
mambocab: :+1:

ccmlib/node.py
```diff
@@ -1427,6 +1427,16 @@ def __update_envfile(self):
...
+            else:
+                gc_log_setting = 'JVM_OPTS="$JVM_OPTS -Xloggc:{}/logs/gc.log"'.format(self.get_path())
```
mambocab: Since the `' -Xloggc:{}/logs/gc.log"'.format(self.get_path())'` part of both of these strings is identical, part of me wants to ask you to pull that out so you only write the `-Xloggc` option part once and there's less danger of incorrectly changing one but not the other in the future. That may be a silly request though; I trust your judgement.

mambocab:
After you address [this comment](https://github.com/pcmanus/ccm/pull/438#discussion_r49203600) (whether you make any changes or not), I'm +1.

PR #440, opened 2016-01-21
Dse fixes

ptnapoleon:
@kishkaru and @mambocab to review, please. CI is running on this now, as far as C* goes.

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -318,10 +318,6 @@ class Cluster(object):
                     if old_node is not node:
                         old_node.watch_log_for_alive(node, from_mark=mark)
 
-        if wait_for_binary_proto:
-            for node, p, mark in started:
-                node.wait_for_binary_interface(process=p, verbose=verbose, from_mark=mark)
-
         return started
 
     def stop(self, wait=True, gently=True):
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -318,6 +318,10 @@ class Cluster(object):
                     if old_node is not node:
                         old_node.watch_log_for_alive(node, from_mark=mark)
 
+        if wait_for_binary_proto:
+            for node, p, mark in started:
+                node.wait_for_binary_interface(process=p, verbose=verbose, from_mark=mark)
+
         return started
 
     def stop(self, wait=True, gently=True):
--- a/ccmlib/dse_node.py
+++ b/ccmlib/dse_node.py
@@ -204,7 +204,37 @@ class DseNode(Node):
             self._stop_agent()
         return stopped
 
-    def dsetool(self, cmd):
+    def nodetool(self, cmd, username=None, password=None, capture_output=True, wait=True):
+        """
+        Setting wait=False makes it impossible to detect errors,
+        if capture_output is also False. wait=False allows us to return
+        while nodetool is still running.
+        """
+        if capture_output and not wait:
+            raise common.ArgumentError("Cannot set capture_output while wait is False.")
+        env = self.get_env()
+        nodetool = common.join_bin(self.get_install_dir(), 'bin', 'nodetool')
+        args = [nodetool, '-h', 'localhost', '-p', str(self.jmx_port)]
+        if username is not None:
+            args += [ '-u', username]
+        if password is not None:
+            args += [ '-pw', password]
+        args += cmd.split()
+        if capture_output:
+            p = subprocess.Popen(args, env=env, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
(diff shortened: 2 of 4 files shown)


ccmlib/cluster.py
```diff
@@ -318,10 +318,6 @@ def start(self, no_wait=False, verbose=False, wait_for_binary_proto=False, wait_
                     if old_node is not node:
                         old_node.watch_log_for_alive(node, from_mark=mark)
 
-        if wait_for_binary_proto:
```
ptnapoleon: This worries me, and I think I'd like to undo this change, assuming DSE still works w/o it.

ccmlib/dse_node.py
```diff
@@ -66,8 +66,8 @@ def start(self,
               wait_other_notice=False,
               replace_token=None,
               replace_address=None,
-              jvm_args=None,
-              wait_for_binary_proto=False,
+              jvm_args=[],
```
ptnapoleon: This jvm_args change is clearly wrong and needs fixed.

ccmlib/node.py
```diff
@@ -775,13 +776,18 @@ def run_cqlsh(self, cmds=None, show_output=False, cqlsh_options=None, return_out
             for cmd in cmds.split(';'):
                 cmd = cmd.strip()
                 if cmd:
-                    cmd_str += (cmd + ';\n')
-            cmd_str += "quit;\n"
-            output = p.communicate(input=cmd_str)
+                    p.stdin.write(cmd + ';\n')
+            p.stdin.write("quit;\n")
+            p.wait()
```
ptnapoleon: This line change doesn't seem right, as now we no longer have `output` defined.

ptnapoleon:
Okay this is ready for review from the two of you. I made some changes, as this broke some dtests.

kishkaru:
Works great, tested on latest 4.7 and 4.8.

ptnapoleon:
Cassci just got back to me. Looks like this is failing some dtests.

ccmlib/dse_node.py
```diff
@@ -67,7 +67,7 @@ def start(self,
               replace_token=None,
               replace_address=None,
               jvm_args=None,
-              wait_for_binary_proto=False,
+              wait_for_binary_proto=True,
```
mambocab: Worried about this API change; is it necessary? I don't know anything about pcmanus/ccm's DSE userbase, fwiw.

ccmlib/dse_node.py
```diff
@@ -95,6 +95,8 @@ def start(self,
         if wait_other_notice:
             marks = [(node, node.mark_log()) for node in list(self.cluster.nodes.values()) if node.is_running()]
 
+        self.mark = self.mark_log()
```
mambocab: Any particular reason to make this `self.mark` and not just `mark`? I worry about random state on a started node being an attractive nuisance.

ccmlib/dse_node.py
```diff
@@ -180,7 +178,10 @@ def start(self,
                 node.watch_log_for_alive(self, from_mark=mark)
 
         if wait_for_binary_proto:
-            self.wait_for_binary_interface()
+            self.watch_log_for("Starting listening for CQL clients", from_mark=self.mark)
```
mambocab: Are we moving away from `wait_for_binary_interface` so we can use the mark? I'd rather add a `mark` parameter to `wait_for_binary_interface` than duplicate the `"Starting listening for CQL clients"`. Alternatively we could define that string as a shared constant somewhere.

mambocab:
Seems reasonable outside my comments.

ccmlib/dse_node.py
ptnapoleon: It already has a mark parameter, afaik. I believe it may have been to stop checking if the socket is listening.

ccmlib/dse_node.py
ptnapoleon: It's self.mark in node.py currently. We can hash out the fix on the main branch there, but symmetry is preferred for now.

ccmlib/node.py
```diff
@@ -378,7 +378,7 @@ def watch_log_for(self, exprs, from_mark=None, timeout=600, process=None, verbos
                     if process.returncode != 0:
                         raise RuntimeError()  # Shouldn't reuse RuntimeError but I'm lazy
 
-        with open(log_file) as f:
+        with open(self.logfilename()) as f:
```
ptnapoleon: This change is what broke all of the tests.

ptnapoleon:
Okay, works for launching DSE, and I have responded to all of Jim's comments. Assuming CI is happy with the latest run, I will merge this.

PR #445, opened 2016-02-02
feature: add a flag to enable PasswordAuthenticator

thibaultCha:
This adds a flag to change the authenticator to `PasswordAuthenticator` (with default credentials). It comes handy when doing unit testing of Cassandra drivers that need to test their implementation of the Plain text authenticator flow.

If #431 is accepted this could potentially be obsolete, though it can also be considered as syntactic sugar.


ptnapoleon:
+1

PR #448, opened 2016-02-05
Stagger node starts in jdk7

ptnapoleon:
@mambocab or @jkni to review

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -284,6 +284,14 @@ class Cluster(object):
                     mark = node.mark_log()
 
                 p = node.start(update_pid=False, jvm_args=jvm_args, profile_options=profile_options, verbose=verbose, quiet_start=quiet_start, allow_root=allow_root)
+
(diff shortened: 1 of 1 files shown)


ccmlib/cluster.py
```diff
@@ -284,6 +284,14 @@ def start(self, no_wait=False, verbose=False, wait_for_binary_proto=False, wait_
                     mark = node.mark_log()
 
                 p = node.start(update_pid=False, jvm_args=jvm_args, profile_options=profile_options, verbose=verbose, quiet_start=quiet_start, allow_root=allow_root)
+
+                # Prior to JDK8, starting every node at once could lead to a
+                # nanotime collision where the RNG that generates a node's tokens
+                # gives identical tokens to several nodes. Thus, we stagger
+                # the node starts
+                if common.get_jdk_version() < '1.8':
+                    time.sleep(1)
```
mambocab: You can make this sleep way lower -- we just need to get different random values.

mambocab:
Other than my comment on the length of the sleep, I'm +1.

PR #450, opened 2016-02-09
Support for graph and dsefs for DSE 5.0 EAP1

ptnapoleon:


ptnapoleon:
+1. Thanks @mike-tr-adamson

PR #455, opened 2016-02-13
support for proactive log watching

knifewine:


knifewine:
for usage example see https://github.com/riptano/cassandra-dtest/pull/809

ccmlib/cluster.py
```diff
@@ -100,6 +102,39 @@ def set_install_dir(self, install_dir=None, version=None, verbose=False):
...
+        def grep_and_mark_all_logs():
+            errordata = OrderedDict()
+
+            try:
+                for node in self.nodelist():
+                    errors = node.grep_log_for_errors_from(seek_start=log_positions[node.name])
+                    log_positions[node.name] = node.mark_log()
+                    if errors:
+                        errordata[node.name] = errors
+            except IOError, e:
```
ptnapoleon: Can you verify this is legal Python 2 and Python 3?

ccmlib/cluster.py
```diff
@@ -100,6 +102,39 @@ def set_install_dir(self, install_dir=None, version=None, verbose=False):
...
+                if e.strerror == 'No such file or directory':
```
ptnapoleon: Shouldn't this be `if 'No such file or directory' in e.strerror:`?

knifewine: good idea, pushed.

ccmlib/cluster.py
knifewine: fixed/pushed.

ccmlib/cluster.py
```diff
@@ -100,6 +102,39 @@ def set_install_dir(self, install_dir=None, version=None, verbose=False):
...
+                if 'No such file or directory' in e.strerror:
+                    pass  # most likely log file isn't yet written
+                else:
+                    raise
+
+            if errordata:
+                on_error_call(errordata)
+
+            timer = threading.Timer(interval, grep_and_mark_all_logs)
+            timer.daemon = True  # so it will exit when the main process exits
```
ptnapoleon: Are you sure this comment matches behavior? I'm not drawing that from the docs, but maybe I'm misreading them?
https://docs.python.org/2/library/threading.html#threading.Thread

knifewine: down a bit, here: https://docs.python.org/2/library/threading.html#threading.Thread.daemon

knifewine: this is basically to protect against the thread continuing in a rogue way after the main python process is going. esp important since the error handling in the thread is a bit liberal to keep it from dying unnecessarily.

knifewine: terribly worded in the docs, but this is it "the entire Python program exits when no alive non-daemon threads are left."
meaning, the entire Python program (including threads) exits when no alive non-daemon threads (e.g. the main program) are left.

ptnapoleon: Ahh, that double negative threw me. I was reading it backwards.

ptnapoleon:
+1

PR #457, opened 2016-02-26
Only update the envfile in __update_envfile

ptnapoleon:
@mambocab to review, while I run on CI. This fixes the relevant windows test, which is promising.

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -185,19 +185,11 @@ def make_cassandra_env(install_dir, node_path, update_conf=True):
     if not os.path.exists(dst):
         shutil.copy(orig, dst)
 
-    if update_conf:
-        replacements = ""
-        if is_win() and get_version_from_build(node_path=node_path) >= '2.1':
-            replacements = [
-                ('env:CASSANDRA_HOME =', '        $env:CASSANDRA_HOME="%s"' % install_dir),
-                ('env:CASSANDRA_CONF =', '    $env:CCM_DIR="' + node_path + '\\conf"\n    $env:CASSANDRA_CONF="$env:CCM_DIR"'),
-                ('cp = ".*?env:CASSANDRA_HOME.conf', '    $cp = """$env:CASSANDRA_CONF"""')
-            ]
-        else:
-            replacements = [
-                ('CASSANDRA_HOME=', '\tCASSANDRA_HOME=%s' % install_dir),
-                ('CASSANDRA_CONF=', '\tCASSANDRA_CONF=%s' % os.path.join(node_path, 'conf'))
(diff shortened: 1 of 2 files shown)


ptnapoleon:
This breaks every single dtest, hold off while I rewrite the patch

ptnapoleon:
Okay, review again, please. Running on CI but it appears to work now.

ptnapoleon:
CI says this is worth reviewing. Thoughts @mambocab?

ccmlib/common.py
```diff
@@ -185,19 +185,11 @@ def make_cassandra_env(install_dir, node_path, update_conf=True):
...
-                ('env:CASSANDRA_HOME =', '        $env:CASSANDRA_HOME="%s"' % install_dir),
-                ('env:CASSANDRA_CONF =', '    $env:CCM_DIR="' + node_path + '\\conf"\n    $env:CASSANDRA_CONF="$env:CCM_DIR"'),
-                ('cp = ".*?env:CASSANDRA_HOME.conf', '    $cp = """$env:CASSANDRA_CONF"""')
-            ]
-        else:
-            replacements = [
-                ('CASSANDRA_HOME=', '\tCASSANDRA_HOME=%s' % install_dir),
-                ('CASSANDRA_CONF=', '\tCASSANDRA_CONF=%s' % os.path.join(node_path, 'conf'))
-            ]
+    if update_conf and not (is_win() and get_version_from_build(node_path=node_path) >= '2.1'):
```
mambocab: This could do with some documenting comments and/or pulling out the guard logic into a variable.

ccmlib/node.py
```diff
@@ -1417,6 +1417,16 @@ def __update_envfile(self):
 
         common.replace_in_file(conf_file, jmx_port_pattern, jmx_port_setting)
 
+
+        if common.is_win() and common.get_version_from_build(node_path=self.get_path()) >= '2.1':
```
mambocab: Same here -- there should be some documentation about the changes in C* that make this necessary.

mambocab:
Re: my last 2 comments: since those use the same guard and has a very specific meaning, maybe make it a method on `Node`, like `node.cassandra_env_in_bin_dir()`. There are conditions where env vars are set somewhere special, and I think both of these bits of code would be clearer if you could explicitly ask the node that.

mambocab:
If this patch makes stuff work again, merge away. I don't understand the need for these methods well enough to give a really confident +1, but that's expected.

ptnapoleon:
Filed a cstar ticket to clean up the windows gating logic everywhere. Merging this since it's a horribly critical bug.

PR #461, opened 2016-03-07
Improve handling when not all loopback addresses are available

bcantoni:
On Mac the user needs to make sure the loopback addresses (127.0.0.2, 127.0.0.3, etc.)
are created. Currently if they aren't created, ccm fails when creating the cluster,
but leaves it in a partial state (usually just the first node is running).

This change does a pre-flight check to make sure all loopbacks are available, print
a better warning, and exit before starting any nodes.

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -5,6 +5,7 @@ import os
 import random
 import shutil
 import subprocess
+import sys
 import threading
 import time
 from collections import OrderedDict, defaultdict
@@ -313,6 +314,15 @@ class Cluster(object):
             jvm_args = []
 
         common.assert_jdk_valid_for_cassandra_version(self.cassandra_version())
+
+        # check whether all loopback aliases are available before starting any nodes
+        for node in self.nodes.values():
+            for itf in node.network_interfaces.values():
+                if itf is not None:
+                    if not common.check_socket_available(itf, return_on_error=True):
+                        addr, port = itf
+                        print_("Inet address %s:%s is not available; a cluster may already be running or you may need to add the loopback alias" % (addr, port))
+                        sys.exit(1)
 
         started = []
         for node in list(self.nodes.values()):
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -395,7 +395,7 @@ def validate_install_dir(install_dir):
         raise ArgumentError('%s does not appear to be a cassandra or dse installation directory' % install_dir)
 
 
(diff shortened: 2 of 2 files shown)


ccmlib/cluster.py
```diff
@@ -314,6 +315,15 @@ def start(self, no_wait=False, verbose=False, wait_for_binary_proto=False,
 
         common.assert_jdk_valid_for_cassandra_version(self.cassandra_version())
 
+        # check whether all loopback aliases are available before starting any nodes
+        for node in list(self.nodes.values()):
```
mambocab: I don't think building a list, here or in the line below, is necessary.

ccmlib/common.py
```diff
@@ -395,7 +395,7 @@ def validate_install_dir(install_dir):
         raise ArgumentError('%s does not appear to be a cassandra or dse installation directory' % install_dir)
 
 
-def check_socket_available(itf):
+def check_socket_available(itf, return_on_error=False):
```
mambocab: The "return a bool" interface is much better than the existing one in my opinion. @ptnapoleon you should consider refactoring the uses of this function to make this the default interface.

mambocab:
Other than my little nit about building `list`s, this looks good. I'll let @ptnapoleon do the final review, though, when he's back online.

bcantoni:
@mambocab thanks - I've updated to remove those extra list() and improved the error message a bit; I'll squash after reviews

ccmlib/cluster.py
```diff
@@ -314,6 +315,15 @@ def start(self, no_wait=False, verbose=False, wait_for_binary_proto=False,
 
         common.assert_jdk_valid_for_cassandra_version(self.cassandra_version())
 
+        # check whether all loopback aliases are available before starting any nodes
+        for node in self.nodes.values():
+            for itf in node.network_interfaces.values():
+                if itf is not None:
+                    if not common.check_socket_available(itf, return_on_error=True):
+                        addr, port = itf
+                        print "Inet address %s:%s is not available; a cluster may already be running or you may need to add the loopback alias" % (addr, port)
```
ptnapoleon: You will need to `from six import print_`, and then use the `print_` function here, to maintain compatibility across python 2 and 3.

ptnapoleon:
+1 after the Python2/3 compat fix.

bcantoni:
Review comments addressed and commits squashed - should be ready to go

PR #465, opened 2016-03-14
Restore previous ability of cluster.start() to start partially downed clusters

jkni:
This PR only checks interface availability for nodes that aren't started when doing a cluster.start(). This keeps the positive effects of the previous PR (no partial states if another cluster is left running or interfaces not available) while restoring idempotent starts and the ability to use start for only partially downed clusters.


ptnapoleon:
+1. You're the best.

PR #466, opened 2016-03-15
Add node level command for updatedseconf. Literal yaml for node cmds.

tolbertam:
* Node level command for `updatedseconf`  (i.e. `ccm node1 updatedseconf hive_metastore_enabled:false`)
* Add support for --yaml for node level commands.
* Small doc change for updatedseconf's yaml option.


ptnapoleon:
+1, thanks Andy

PR #470, opened 2016-03-18
Set universal_newlines=True on subprocess.Popen. Closes #469

### LINKED ISSUE #469: Fix Python3 nodetool status output
From #468 

https://github.com/pcmanus/ccm/pull/468#issuecomment-198105583

ptnapoleon:
This lets us get back strings instead of bytes in python3

@kishkaru to review

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -687,7 +687,7 @@ class Node(object):
         args = [nodetool, '-h', 'localhost', '-p', str(self.jmx_port)]
         args += cmd.split()
         if capture_output:
-            p = subprocess.Popen(args, env=env, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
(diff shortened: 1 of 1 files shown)


mambocab:
I'm concerned about impact on windows -- this could get confusing if we, e.g., try to split on `os.linesep`.

The fact that `universal_newlines` changes the return type on Py3 is reeeeal confusing imo. Seems like a poorly-named thing. If we use this, this behavior deserves a comment in our code.

I'm... +0? I think I'd rather explicitly do something like `stdout, stderr = map(str, p.communicate())`, but I don't feel too strongly.

ptnapoleon:
Casting to str does not fix the root issue.

```
x = b'bytes\nstring'
print(str(x))
"b'bytes\nstring'"
```
is what happens

mambocab:
Huh, ok. I guess I haven't dealt with text/bytes processing enough on Py3 to know how to do this explicitly. +1 to these changes (having not run them).

PR #472, opened 2016-03-20
Update spark-env.sh with node-specific environment variables.

tolbertam:
Noticed that the environment variables for spark defined in `common.make_dse_env` were not actually being used because `resources/spark/conf/spark-env.sh` was overriding them when it was sourced in.   Fixed this by updating this file with the proper environment variables from any environment variable in `make_dse_env` found in this file.

Without this, files get created in /var/lib/spark and /var/log/spark.


ptnapoleon:
+1

PR #473, opened 2016-03-20
Allow nesting > 2 for updateconf/updatedseconf, persist and read dse_config_options

tolbertam:
This does a bunch of little things which may not be useful in the general case, but is good to have:

* Previously with updateconf/updatedseconf, someone could specify a nested structure, i.e.:

    ```bash
    ccm updateconf "client_encryption_options.enabled: true"
    ```

    This would update cassandra.yaml with:

    ```bash
    client_encryption_options:
      enabled: true
    ```

    In the general case this is more than enough, but there were a few problems:

    1. It would update in place for updateconf (i.e. if there are other options under client_encryption_options, those would remain unchanged), but not with updatedseconf (it'd blow away all children except the newly added element), reconciled that logic into `common#merge_configuration`.
    2. You couldn't nest > 2 deep, i.e. "spark_application_info_options.driver.sink:true" would not update:

        ```yaml
        spark_application_info:
          driver:
            sink: true
        ```

        Instead it would do:

        ```yaml
        spark_application_info.driver.sink: true
        ```

    Commit 37afe6a reconciles this.

* Configuration for updatedseconf at both the cluster and node level was not being read in when loading the cluster/nodes.   It was also not being persisted at the node level (my mistake since I added that capability).  Commit 201694f reconciles this.

    An example to experiment with this:

    Update and cluster level with a delete, and then the node level undoing the delete with nested values:

    ```bash
    ccm updatedseconf "spark_application_info_options.driver.sink:true" "spark_application_info_options.driver.jvmSource:"
    ccm node1 updatedseconf "spark_application_info_options.driver.jvmSource:true"
    ```

    sink should be true in dse.yaml for all nodes, jvmSource should be present in node1, but not the rest of the nodes.

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -2,6 +2,7 @@
 # Cassandra Cluster Management lib
 #
 
+import copy
 import fnmatch
 import os
 import platform
@@ -481,12 +482,22 @@ def parse_settings(args, literal_yaml=False):
                 except ValueError:
                     pass
             splitted = key.split('.')
-            if len(splitted) == 2:
-                try:
-                    settings[splitted[0]][splitted[1]] = val
-                except KeyError:
-                    settings[splitted[0]] = {}
-                    settings[splitted[0]][splitted[1]] = val
+            split_length = len(splitted)
+            if split_length >= 2:
+                # Where we are currently at in the dict.
+                tree_pos = settings
+                # Iterate over each split and build structure as needed.
+                for pos in xrange(split_length):
+                    split = splitted[pos]
+                    if pos == split_length - 1:
+                        # If at the last split, set value.
+                        tree_pos[split] = val
+                    else:
+                        # If not at last split, create a new dict at the current
+                        # position for this split.
+                        if split not in tree_pos:
+                            tree_pos[split] = {}
+                        tree_pos = tree_pos[split]
             else:
                 settings[key] = val
     return settings
@@ -591,3 +602,26 @@ def assert_jdk_valid_for_cassandra_version(cassandra_version):
     if cassandra_version >= '3.0' and get_jdk_version() < '1.8':
         print_('ERROR: Cassandra 3.0+ requires Java >= 1.8, found Java {}'.format(get_jdk_version()))
         exit(1)
+
+
+def merge_configuration(original, changes, delete_empty=True):
+    if not isinstance(original, dict):
+        # if original is not a dictionary, assume changes override it.
+        new = changes
+    else:
+        # Copy original so we do not mutate it.
+        new = copy.deepcopy(original)
+        for k, v in changes.iteritems():
+            # If the new value is None or an empty string, delete it
+            # if its in the original data.
+            if delete_empty and k in new and \
+                    (v is None or (isinstance(v, str) and len(v) == 0)):
+                del new[k]
+            else:
+                new_value = v
+                # If key is in both dicts, update it with new values.
+                if k in new:
+                    if isinstance(v, dict):
+                        new_value = merge_configuration(new[k], v, delete_empty)
+                new[k] = new_value
+    return new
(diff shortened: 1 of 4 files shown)


ccmlib/common.py
```diff
@@ -591,3 +602,26 @@ def assert_jdk_valid_for_cassandra_version(cassandra_version):
     if cassandra_version >= '3.0' and get_jdk_version() < '1.8':
         print_('ERROR: Cassandra 3.0+ requires Java >= 1.8, found Java {}'.format(get_jdk_version()))
         exit(1)
+
+
+def merge_configuration(original, changes, delete_empty=True):
```
tolbertam: This is recursive, so could blow stack with arbitrarily deep config, but figure that is ok as I wouldn't expect dse/cassandra yaml to be very nested.

ptnapoleon:
I'm running dtest with your changes, to confirm it breaks nothing there.

tolbertam:
> I'm running dtest with your changes, to confirm it breaks nothing there.

Cool, thanks!  Definitely will be nice to exercise it :)

dins2k2:
Hello, How to update the seed_provider and the respective seeds IPs of a node through updadeconf command?

tolbertam:
> Hello, How to update the seed_provider and the respective seeds IPs of a node through updadeconf command?

You will need to use the --yaml/-y option to pass in literal yaml to accomplish this, this would do it:

```bash
ccm updateconf -y "seed_provider: [{class_name: org.apache.cassandra.locator.SimpleSeedProvider, parameters: [seeds: '8.6.7.5,1.2.3.4']}]"
```

ccmlib/common.py
```diff
@@ -481,12 +482,22 @@ def parse_settings(args, literal_yaml=False):
...
-                    settings[splitted[0]][splitted[1]] = val
-                except KeyError:
-                    settings[splitted[0]] = {}
-                    settings[splitted[0]][splitted[1]] = val
+            split_length = len(splitted)
+            if split_length >= 2:
+                # Where we are currently at in the dict.
+                tree_pos = settings
+                # Iterate over each split and build structure as needed.
+                for pos in xrange(split_length):
```
ptnapoleon: This needs to be `range`, not `xrange`

ccmlib/common.py
ptnapoleon: 100% fine with me.

ccmlib/common.py
```diff
@@ -591,3 +602,26 @@ def assert_jdk_valid_for_cassandra_version(cassandra_version):
...
+                    (v is None or (isinstance(v, str) and len(v) == 0)):
+                del new[k]
+            else:
+                new_value = v
+                # If key is in both dicts, update it with new values.
+                if k in new:
+                    if isinstance(v, dict):
+                        new_value = merge_configuration(new[k], v, delete_empty)
+                new[k] = new_value
+    return new
```
ptnapoleon: Can you restore the new line at the end of the file?

ccmlib/common.py
```diff
@@ -591,3 +602,26 @@ def assert_jdk_valid_for_cassandra_version(cassandra_version):
...
+def merge_configuration(original, changes, delete_empty=True):
+    if not isinstance(original, dict):
+        # if original is not a dictionary, assume changes override it.
+        new = changes
+    else:
+        # Copy original so we do not mutate it.
+        new = copy.deepcopy(original)
+        for k, v in changes.iteritems():
+            # If the new value is None or an empty string, delete it
+            # if its in the original data.
```
ptnapoleon: nit: `it's`

ptnapoleon:
After the `xrange` --> `range` fix, I'm good to merge.

tolbertam:
Thanks for reviewing!  I've addressed your feedback in ea25a6c

PR #475, opened 2016-03-21
allow flush to wait if needed

knifewine:


ptnapoleon:
-1. The default wait value when calling nodetool directly is True. I don't want different behaviors between 

`node.nodetool('flush')`
and
`node.flush()`

knifewine:
oh yup, sorry, I misread when I checked there earlier.

knifewine:
since that's the case, there's no need for this anyway. closing.

PR #476, opened 2016-03-21
Fixing ipv6 for C* with versions > 3.2

GregBestland:
This allows me to spin up C* cluster newer the 3.2 with IPV6.

I tested these changes against 3.4, and 3.0.3  It seems wot work with both paths. 
Take a look and let me know. I'm not sure the "-Djava.net.preferIPv6Addresses=true" was needed as it worked fine without it so I left it absent, and simply removed '-Djava.net.preferIPv4Stack=true' from the jvm.options file


ptnapoleon:
Well, it lgtm. I'm going to run with dtest to make sure there are no unexpected side effects and then merge. Thanks @GregBestland!

PR #479, opened 2016-03-23
Assert that wait_other_notice and join_ring are not used together

michaelsembwever:
Simple assertion that wait_other_notice and join_ring are not used together.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -512,6 +512,8 @@ class Node(object):
                 common.check_socket_available(itf)
 
         if wait_other_notice:
+            if not join_ring:
+                raise NodeError("Other nodes will never see this node %s as UP when join_ring=False" % self.name)
             marks = [(node, node.mark_log()) for node in list(self.cluster.nodes.values()) if node.is_live()]
 
         self.mark = self.mark_log()


ptnapoleon:
+1, thanks!

ptnapoleon:
Hey @michaelsembwever, it doesn't seem that your assumption is correct. At least with trunk, I can start a node with join_ring=False, and other nodes WILL notice it is UP. Where are you seeing the behavior that prompted this? 

This is also breaking a number of dtests, so I'm going to revert this PR until we figure it out.

michaelsembwever:
@ptnapoleon that's odd.

When I was writing the dtest for https://issues.apache.org/jira/browse/CASSANDRA-11381 it failed when using `node.start(wait_other_notice=True, join_ring=False)`

so this assertion made sense. but i guess i've missed something…

michaelsembwever:
> This is also breaking a number of dtests…

got a link to them? would be interesting to see who's doing this.

ptnapoleon:
These failures list the tests that do so
http://cassci.datastax.com/job/trunk_dtest/1076/testReport/

michaelsembwever:
Didn't help i was working with dtests branch `trunk` instead of `master` :-/

ptnapoleon:
Oh man, is that branch still around/? My bad, I will delete.

michaelsembwever:
When a node comes up with `join_ring=false` the other nodes will **not** report it as `UP`.
Those tests that failed were correct in failing. Although it is a programmatic failure. 

I can also confirm that the dtest in https://issues.apache.org/jira/browse/CASSANDRA-11381 does not work with a combination of `(wait_other_notice=True, join_ring=False)`.

I've put in a fix for the dtests that failed. see https://github.com/riptano/cassandra-dtest/pull/892

I don't know why `bootstrap_test.py` failed. IWFM.

jkni:
I think this is one of those situations where everyone is "right".

The behavior of join_ring is slightly different depending on whether the node has saved tokens. If there are no saved tokens, we never set a status, so the status in gossip is null, which isn't a "dead" status, so handleMajorStateChange marks the node alive, which results in the log message for which wait_other_notice looks.

If the node has saved tokens, we set the status to hibernate, which is a "dead" status, so handleMajorStateChange doesn't mark the node alive, which means that the log message that for which wait_other_notice looks doesn't appear.

I haven't thought too much about if there is a way to make this more normalized on the C* side - on the CCM side, I think we could check if the node is [uninitialized](https://github.com/pcmanus/ccm/blob/master/ccmlib/node.py#L28) and disallow the combination of any initialized status with join_ring=False and wait_other_notice=True, but I haven't made an effort to really think that through yet.

PR #482, opened 2016-03-24
add node.Node.envfilename method

mambocab:
[This code](https://github.com/beobal/cassandra-dtest/blob/10091/jmxutils.py#L55) from [@beobal's recent PR](https://github.com/riptano/cassandra-dtest/pull/887) is necessary but silly. This PR adds a method that makes it unnecessary.

(Also, if it weren't inconsistent with existing APIs, I'd make this a `@property`. Is that a refactoring we'd consider in the long term? If so, maybe we could make this one from the get-go.)


ptnapoleon:
+1

PR #485, opened 2016-03-29
Fixed memory label to align to CASSANDRA-9692 

jkni:
[CASSANDRA-9692](https://issues.apache.org/jira/browse/CASSANDRA-9692) unifies/improves logging/output of data sizes and rates throughout C*.

As part of this normalization, it changes the output of nodetool info to use the correct SI prefixes. The only dependency on this I could locate for CCM is in its parsing of the nodetool info output. This parsing happens in a function without the C* version available, so we can't do a version check to determine correct units. This PR proposes the easy route of supporting both unit types; another option is to refactor the tests/call sites to pass in a C* version, so that we can decide based on version.

(Please note that I'm reviewer for the above ticket, not assignee. PRing this since it is a new contributor and managing a C* ticket/CCM PR/dtest PR isn't the easiest thing).

### DIFF
--- a/tests/test_lib.py
+++ b/tests/test_lib.py
@@ -177,30 +177,55 @@ class TestRunCqlsh(ccmtest.Tester):
 class TestNodeLoad(ccmtest.Tester):
 
     def test_rejects_multiple_load_lines(self):
+        info = 'Load : 699 KiB\nLoad : 35 GiB'
+        with self.assertRaises(RuntimeError):
+            ccmlib.node._get_load_from_info_output(info)
+
         info = 'Load : 699 KB\nLoad : 35 GB'
         with self.assertRaises(RuntimeError):
             ccmlib.node._get_load_from_info_output(info)
 
     def test_rejects_unexpected_units(self):
-        infos = ['Load : 200 PB', 'Load : 12 Parsecs']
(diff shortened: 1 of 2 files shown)


tests/test_lib.py
```diff
@@ -177,24 +177,22 @@ def test_print_and_return_output(self):
...
+        infos = ['Load : 200 PiB', 'Load : 12 Parsecs']
 
         for info in infos:
             with self.assertRaises(RuntimeError):
                 ccmlib.node._get_load_from_info_output(info)
 
     def test_gets_correct_value(self):
-        info_value = [('Load : 328.45 KB', decimal.Decimal('328.45')),
-                      ('Load : 295.72 MB', decimal.Decimal('295.72') * 1024),
-                      ('Load : 183.79 GB',
```
mambocab: We still want correct handling of `GB` on versions that use it, so could you add tests for the `GiB` behavior, rather than replacing these? (and the same for other `X{,i}B` units)

mambocab: (I realize that this means asking you to fix tests that were broken when you got here. Sorry about that)

jkni: Good idea - done!

mambocab: They look great, +1 from me on the changes to the tests based on visual inspection.

ptnapoleon:
+1

PR #487, opened 2016-04-01
Use sstabledump for json command if available

spodkowinski:
`ccm json` is currently broken for Cassandra 3.x, but we can just use the `sstabledump` tool in this case.


ptnapoleon:
+1

PR #489, opened 2016-04-01
NodetoolError not found

stamhankar999:
We're missing an import in dse_node.py.


ptnapoleon:
I'm happy to merge in the NodetoolError import now, but the bash completion is going to take some time to review. Can you possibly only include the import fix in this PR?

stamhankar999:
Yikes; my intention was to have two PRs, one with each fix. I see now (and understand how) both commits accidentally ended up in this PR. I'll close out this PR and create a new one with just the NodetoolError fix.

PR #499, opened 2016-04-12
Add timeout to (node|cluster).wait_for_compactions()

pauloricardomg:


ccmlib/node.py
```diff
@@ -680,16 +680,20 @@ def stop(self, wait=True, wait_other_notice=False, gently=True):
...
+        elapsed = 0
+        while elapsed < timeout:
             output, err = self.nodetool("compactionstats", capture_output=True)
             if pattern.search(output):
-                break
-            time.sleep(10)
+                return
+            elapsed = elapsed + 1
+            time.sleep(1)
+        raise TimeoutError(time.strftime("%d %b %Y %H:%M:%S", time.gmtime()) + " [" + self.name + "] Compactions did not finish in {} seconds".format(timeout))
```
ptnapoleon: Could you use `.format()` to build this entire string, and not do any concatenation?

ccmlib/node.py
```diff
@@ -680,16 +680,20 @@ def stop(self, wait=True, wait_other_notice=False, gently=True):
...
         pattern = re.compile("pending tasks: 0")
-        while True:
+        elapsed = 0
+        while elapsed < timeout:
             output, err = self.nodetool("compactionstats", capture_output=True)
             if pattern.search(output):
-                break
-            time.sleep(10)
+                return
+            elapsed = elapsed + 1
```
ptnapoleon: Could you make this `elapsed += 1`?

ccmlib/node.py
```diff
@@ -680,16 +680,20 @@ def stop(self, wait=True, wait_other_notice=False, gently=True):
         else:
             return False
 
-    def wait_for_compactions(self):
+    def wait_for_compactions(self, timeout=600):
```
ptnapoleon: Given we aren't tracking seconds, but rather loop iterations, could you make this default smaller?

Or could you actually track time passed?

ccmlib/cluster.py
```diff
@@ -401,14 +401,15 @@ def set_log_level(self, new_level, class_names=None):
...
         """
         Wait for all compactions to finish on all nodes.
         """
+        result = True
         for node in list(self.nodes.values()):
             if node.is_running():
-                node.wait_for_compactions()
-        return self
+                node.wait_for_compactions(timeout)
+        return result
```
ptnapoleon: This never returns anything but True?

pauloricardomg:
thanks for the review, comments addressed.
I left the loop-based implementation and reduced timeout to 120 as that is simplest to to now (its similar to the watch_log_for implementation, so we'll have to update both to make use of proper timeouts, we can do it in a separate ticket)

ptnapoleon:
+1, thanks Paulo

PR #502, opened 2016-04-26
Enable seed addresses to be explicitly set

beobal:
So we can specify that a node's broadcast_address can be used in the seed list. See CASSANDRA-10134 (specifically https://issues.apache.org/jira/browse/CASSANDRA-10134?focusedCommentId=15243094&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15243094)


ptnapoleon:
+1

PR #504, opened 2016-05-26
separated batch_commitlog from set_configuration_options

smccarthy788:
@mambocab, please review this as well.

Connected with https://github.com/riptano/cassandra-dtest/pull/1002

Cassci test: http://cassci.datastax.com/view/Dev/view/smccarthy788/job/smccarthy788-cassandra-3.0-dtest/lastCompletedBuild/

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -444,25 +444,20 @@ class Cluster(object):
             raise common.ArgumentError("No live node")
         livenodes[0].run_cli(cmds, show_output, cli_options)
 
-    def set_configuration_options(self, values=None, batch_commitlog=None):
+    def set_configuration_options(self, values=None):
         if values is not None:
             for k, v in iteritems(values):
                 self._config_options[k] = v
-        if batch_commitlog is not None:
-            if batch_commitlog:
-                self._config_options["commitlog_sync"] = "batch"
-                self._config_options["commitlog_sync_batch_window_in_ms"] = 5
-                self._config_options["commitlog_sync_period_in_ms"] = None
-            else:
-                self._config_options["commitlog_sync"] = "periodic"
-                self._config_options["commitlog_sync_period_in_ms"] = 10000
-                self._config_options["commitlog_sync_batch_window_in_ms"] = None
 
         self._update_config()
         for node in list(self.nodes.values()):
             node.import_config_files()
         self.__update_topology_files()
         return self
+
+    def set_batch_commitlog(self, enabled):
+        for node in list(self.nodes.values()):
+            node.set_batch_commitlog(enabled=enabled)
 
     def set_dse_configuration_options(self, values=None):
         raise common.ArgumentError('Cannot set DSE configuration options on a Cassandra cluster')
--- a/ccmlib/cmds/cluster_cmds.py
+++ b/ccmlib/cmds/cluster_cmds.py
@@ -694,7 +694,9 @@ class ClusterUpdateconfCmd(Cmd):
         parser.add_option('--no-hh', '--no-hinted-handoff', action="store_false",
                           dest="hinted_handoff", default=True, help="Disable hinted handoff")
         parser.add_option('--batch-cl', '--batch-commit-log', action="store_true",
-                          dest="cl_batch", default=False, help="Set commit log to batch mode")
+                          dest="cl_batch", default=None, help="Set commit log to batch mode")
+        parser.add_option('--periodic-cl', '--periodic-commit-log', action="store_true",
+                          dest="cl_periodic", default=None, help="Set commit log to periodic mode")
         parser.add_option('--rt', '--rpc-timeout', action="store", type='int',
                           dest="rpc_timeout", help="Set rpc timeout")
         parser.add_option('-y', '--yaml', action="store_true", dest="literal_yaml", default=False, help="If enabled, treat argument as yaml, not kv pairs. Option syntax looks like ccm updateconf -y 'a: [b: [c,d]]'")
@@ -704,6 +706,10 @@ class ClusterUpdateconfCmd(Cmd):
         Cmd.validate(self, parser, options, args, load_cluster=True)
         try:
             self.setting = common.parse_settings(args, literal_yaml=self.options.literal_yaml)
+            if self.options.cl_batch and self.options.cl_periodic:
+                print_("Can't set commitlog to be both batch and periodic.{}".format(os.linesep))
+                parser.print_help()
+                exit(1)
         except common.ArgumentError as e:
             print_(str(e), file=sys.stderr)
             exit(1)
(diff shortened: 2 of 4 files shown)


ccmlib/cluster.py
```diff
@@ -444,19 +444,26 @@ def run_cli(self, cmds=None, show_output=False, cli_options=None):
             raise common.ArgumentError("No live node")
         livenodes[0].run_cli(cmds, show_output, cli_options)
 
-    def set_configuration_options(self, values=None, batch_commitlog=None):
+    def set_configuration_options(self, values=None):
```
ptnapoleon: We should let this take `**kwargs`, no? So other callers don't need to change? Though we're already changing behavior out from under them, so maybe we should make the error loud

mambocab: I'm pro-loud-breakage

ccmlib/cluster.py
```diff
@@ -444,19 +444,26 @@ def run_cli(self, cmds=None, show_output=False, cli_options=None):
...
-                self._config_options["commitlog_sync_period_in_ms"] = 10000
-                self._config_options["commitlog_sync_batch_window_in_ms"] = None
+
+        self._update_config()
+        for node in list(self.nodes.values()):
+            node.import_config_files()
+        self.__update_topology_files()
+        return self
+
+    def set_batch_commitlog(self, batch_commitlog=False):
```
mambocab: This is confusing default behavior -- `set_batch_commitlog()` will disable it. I don't think we even need a default value; `def set_batch_commitlog(enabled)` is probably a better signature.

ccmlib/cluster.py
```diff
@@ -444,19 +444,26 @@ def run_cli(self, cmds=None, show_output=False, cli_options=None):
...
+        if batch_commitlog:
+            self._config_options["commitlog_sync"] = "batch"
+            self._config_options["commitlog_sync_batch_window_in_ms"] = 5
+            self._config_options["commitlog_sync_period_in_ms"] = None
+        else:
+            self._config_options["commitlog_sync"] = "periodic"
+            self._config_options["commitlog_sync_period_in_ms"] = 10000
+            self._config_options["commitlog_sync_batch_window_in_ms"] = None
 
         self._update_config()
```
mambocab: Any reason not to just call `set_configuration_options` here?

ccmlib/node.py
```diff
@@ -232,29 +232,34 @@ def get_base_cassandra_version(self):
...
-                self.__config_options["commitlog_sync_batch_window_in_ms"] = 5
-                self.__config_options["commitlog_sync_period_in_ms"] = None
-            else:
-                self.__config_options["commitlog_sync"] = "periodic"
-                self.__config_options["commitlog_sync_period_in_ms"] = 10000
-                self.__config_options["commitlog_sync_batch_window_in_ms"] = None
+
+        self.import_config_files()
+
+    def set_batch_commitlog(self, batch_commitlog=False):
```
mambocab: Same comments about default arguments and piggybacking on `set_configuration_options`.

ccmlib/node.py
```diff
@@ -232,29 +232,34 @@ def get_base_cassandra_version(self):
...
+        commitlog (since it requires setting 2 options and unsetting one).
+        """
+        if batch_commitlog:
+            self.__config_options["commitlog_sync"] = "batch"
+            self.__config_options["commitlog_sync_batch_window_in_ms"] = 5
+            self.__config_options["commitlog_sync_period_in_ms"] = None
+        else:
+            self.__config_options["commitlog_sync"] = "periodic"
+            self.__config_options["commitlog_sync_period_in_ms"] = 10000
+            self.__config_options["commitlog_sync_batch_window_in_ms"] = None
```
mambocab: This isn't a complaint about the PR, but about the original code: does this need to be duplicated across `cluster` and `node`? Can this PR also refactor the `cluster` implementation to use the `node` implementation?

mambocab:
Done with pass 1 of review. Lookin' good :+1:, ping me when you've addressed my comments.

smccarthy788:
@mambocab, @ptnapoleon: I think this is the best. We really can't refactor set_config_opts without refactoring a lot of other functions related to config options.

set_batch_commitlog works as we want it to: calling the node implementation for each node.

Some manual testing says things work fine. Waiting on cassci to finish testing the dtest suite.

ccmlib/cmds/cluster_cmds.py
```diff
@@ -693,8 +693,6 @@ def get_parser(self):
         parser = self._get_default_parser(usage, self.description())
         parser.add_option('--no-hh', '--no-hinted-handoff', action="store_false",
                           dest="hinted_handoff", default=True, help="Disable hinted handoff")
-        parser.add_option('--batch-cl', '--batch-commit-log', action="store_true",
```
mambocab: I'm wary of breaking this API -- we have no idea who uses this option. I guess our strategies are to either reimplement it on top of the new implementation, or to remove it and only reimplement if someone complains. Thoughts @ptnapoleon?

mambocab:
Some cursory digging makes it look like the meaty part of `cluster.Cluster.set_configuration_options` is making changes to a `cluster.conf`. I'm showing my ignorance here, but I have no idea what that file does. An explanation would be nice, but not totally necessary for me to review this.

But, I accept that it means that we can't just delegate to `node.Node.set_configuration_options`. I haven't decided whether I like or dislike those two methods having the same name, thought that bridge is pretty well crossed for now.

mambocab:
And I'm reasonably comfortable with just removing that CLI option, but @ptnapoleon is the decision-maker there as far as I'm concerned.

ccmlib/cmds/cluster_cmds.py
ptnapoleon: I'd way rather reimplement it on top of the new implementation. I don't know why we wouldn't do that. I have too many users to possibly exhaustively ask them all, and the drivers team hates when I remove things, and make them put it back [for good reason].

ptnapoleon:
I have to re-review the final changes, but @mambocab at a later time I can walk you through the way this is currently architected, at which point you'll see why we had to change it

smccarthy788:
@ptnapoleon, @mambocab; I have reimplemented the api. It now is two separate flags, one for setting commit log to batch, the other for setting commit log to periodic. Some manual testing shows it works for both node & cluster.

ccmlib/cmds/cluster_cmds.py
```diff
@@ -704,6 +706,10 @@ def validate(self, parser, options, args):
         Cmd.validate(self, parser, options, args, load_cluster=True)
         try:
             self.setting = common.parse_settings(args, literal_yaml=self.options.literal_yaml)
+            if self.options.cl_batch and self.options.cl_periodic is not None:
```
ptnapoleon: Why not `if self.options.cl_batch and self.options.cl_periodic:`?

ptnapoleon:
+1. Squash and I'll merge

ptnapoleon:
Thanks

PR #508, opened 2016-06-08
Cstar 97

smccarthy788:
@ptnapoleon, @mambocab: These changes still need to be tested. I'm creating this PR so we can discuss early on. Thoughts?

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -715,19 +715,33 @@ class Node(object):
         nodetool = self.get_tool('nodetool')
         args = [nodetool, '-h', 'localhost', '-p', str(self.jmx_port)]
         args += cmd.split()
+
+        p, stdout, stderr = self._shell_out(args, capture_output=capture_output, wait=wait)
+
+        return stdout, stderr
+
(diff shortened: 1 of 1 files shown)


ccmlib/node.py
```diff
@@ -715,19 +714,49 @@ def nodetool(self, cmd, capture_output=True, wait=True):
...
         args += cmd.split()
-        if capture_output:
-            p = subprocess.Popen(args, env=env, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True)
-            stdout, stderr = p.communicate()
+
+        p, stdout, stderr = self._shell_out(args, capture_output=capture_output, wait=wait)
+
+        return stdout, stderr
+
+    def _shell_out(self, cmd, args=None, wait=True, **kwargs):
```
ptnapoleon: I have a very large bikeshed that I'd like to put this method name in.

ptnapoleon:
You don't appear to have simplified any of the APIs for the various tools methods, only abstracted away the Popen

PR #509, opened 2016-06-08
Update sstablemetadata to pass in all sstables at once

carlyeks:
Before, this was running a separate process for each sstable. This changes the return, but also allows us to run only a single process.


ptnapoleon:
+1

PR #510, opened 2016-06-09
Cstar 97

smccarthy788:
@ptnapoleon, @mambocab: I started with nodetool, making the assumption capture_output & wait will always be true. If we capture output we let callers use or ignore stdout / stderr & check the return code. Waiting lets us check for errors. 

All of the other changes follow along generally like this: create the process, communicate, return stdout, stderr, and rc.

The return values will be (stdout, stderr, and returncode) (except for nodetool, which checks rc internally). Return values should be placed inside a tuple to make it easy for callers to ignore some. I like these more than returning the process object from Popen().

Some methods (e.g. run_sstablesplit, run_sstablemetadata, etc) create a number of different subprocesses. The resulting (stdout, stderr, rc) tuples are packaged into a result list, and returned.

Thoughts?

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -1,6 +1,7 @@
 # ccm node
 from __future__ import with_statement
 
+from collections import namedtuple
 from distutils.version import LooseVersion
 import errno
 import glob
@@ -44,7 +45,7 @@ class TimeoutError(Exception):
         Exception.__init__(self, str(data))
 
 
-class NodetoolError(Exception):
+class ToolError(Exception):
 
     def __init__(self, command, exit_status, stdout=None, stderr=None):
         self.command = command
@@ -52,12 +53,12 @@ class NodetoolError(Exception):
         self.stdout = stdout
         self.stderr = stderr
 
-        message = "Nodetool command '%s' failed; exit status: %d" % (command, exit_status)
+        message = "Subprocess {} exited with non-zero status; exit status: {}".format(command, exit_status)
         if stdout:
-            message += "; stdout: "
+            message += "; \nstdout: "
             message += stdout
         if stderr:
-            message += "; stderr: "
+            message += "; \nstderr: "
             message += stderr
 
         Exception.__init__(self, message)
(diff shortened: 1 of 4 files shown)


smccarthy788:
bump @mambocab, @ptnapoleon

ptnapoleon:
Waiting on Jim to see if he likes the new API

ccmlib/node.py
```diff
@@ -1043,30 +1022,33 @@ def run_sstablerepairedset(self, set_repaired=True, datafiles=None, keyspace=Non
         sstablerepairedset = common.join_bin(cdir, os.path.join('tools', 'bin'), 'sstablerepairedset')
         env = self.get_env()
         sstablefiles = self.__gather_sstables(datafiles, keyspace, column_families)
+        results = []
 
         for sstable in sstablefiles:
```
mambocab: Don't block merge on this, but it's worth checking to see if `sstablerepairedset` can take multiple tables. If it can, it'd be nice to only shell out once.

mambocab:
I'm ok with deprecating the `capture_output` part of the API. I'm wary of assuming users always want `wait=True` -- though I'm not sure why. I'm comfortable doing so for `dtest`, and `ccm nodetool <command>` will block like `<command>` would, and that's fine. I might just be cringing at API breakage.

I proposed an API returning the process object because it gives decisions like "what do we do with `stdout` and `stderr`?", "do we care about return codes?", "do we raise an exception on failure?", and "do we block on completion?" to the caller, without requiring us to implement parameters giving callers control over all those. However, if the answer to those questions is something like "we always want `stdout` and `stderr`, we never raise an exception on failure, and we let the caller deal with return codes", then this new API is good.

This might be a case to use a layered API -- factor out methods that return the process object, and implement the proposed `(out, err, rc)` API on top of those.

It'd be nice to make the `nodetool` API consistent with the rest, though that's a big change to make and a heavily-used method.

Thoughts on the above @ptnapoleon? I don't have a hard "like" or "dislike" for the API yet.

mambocab:
Also, this is a big ol' API change, and it'd be nice for this to be blocked on some proposed improvements to dependency management so this doesn't break every dev branch when we merge.

ptnapoleon:
+1 to the layered API idea, that solves every issue I had with both proposed APIs. Brilliant

ptnapoleon:
This needs rebased. Is it ready for review, after that? I think we set it aside, right?

smccarthy788:
I've almost finished & tested the changes. Once that is done I'll go through dtests and make sure everything is using the new api.

ccmlib/node.py
```diff
@@ -1897,3 +1938,13 @@ def log_line_category(line):
                     break
 
     return matches
+
+
+def handle_external_tool_process(process, cmd_args):
+    (out, err) = process.communicate()
```
ptnapoleon: No need for the parens around `out, err`

ccmlib/node.py
```diff
@@ -44,23 +44,23 @@ def __init__(self, data):
...
+        message = "Subprocess {} exited with non-zero status; exit status: {}".format(command, exit_status)
         if stdout:
             message += "; stdout: "
             message += stdout
         if stderr:
             message += "; stderr: "
             message += stderr
 
-        Exception.__init__(self, message)
+        Exception.__init__(self)
```
ptnapoleon: Why remove `message`?

ccmlib/node.py
```diff
@@ -820,30 +813,32 @@ def run_cqlsh(self, cmds=None, show_output=False, cqlsh_options=None, return_out
...
-            output = p.communicate(input=cmd_str)
 
-            for err in output[1].split('\n'):
-                print_("(EE) ", err, end='')
+        return p
+
+    def run_cqlsh(self, cmds=None, cqlsh_options=None):
+        p = self.run_cqlsh_process(cmds, cqlsh_options)
+
+        cmd_str = ''
```
ptnapoleon: Why is this not using `handle_external_process`

ptnapoleon:
All of the methods that had any keyword arguments removed, need `**kwargs` at the end of their header, to preserve backwards compatibility from the callsites

mambocab:
> All of the methods that had any keyword arguments removed, need **kwargs at the end of their header, to preserve backwards compatibility from the callsites

If I understand you correctly here, I'm -1 -- I'd rather have calls that depend on old signatures break loudly. Do I understand you, @ptnapoleon?

ptnapoleon:
Hmm. I agree, but bump the version to `2.2.0` then

smccarthy788:
Running a cassci test here: http://cassci.datastax.com/view/Dev/view/smccarthy788/job/smccarthy788-cassandra-3.0-dtest/19/

ptnapoleon:
Do you have a new CI run? Do we need a corresponding dtest PR?

smccarthy788:
Latest CI build is http://cassci.datastax.com/view/Dev/view/smccarthy788/job/smccarthy788-cassandra-3.0-dtest/23/

Still having some problems. Corresponding dtest branch is: https://github.com/riptano/cassandra-dtest/tree/cstar-97

Haven't made a PR for it yet.

ptnapoleon:
OK, I will resume ignoring this until those are done?

smccarthy788:
@ptnapoleon, can you please review the Cassci build here: http://cassci.datastax.com/view/Dev/view/smccarthy788/job/smccarthy788-cassandra-3.0-dtest/27/testReport/

ptnapoleon:
Done, pinged offline

ptnapoleon:
Running here: http://cassci.datastax.com/view/Dev/view/ptnapoleon/job/ptnapoleon-trunk-dtest/22/

ptnapoleon:
So it took like 18 tries, but I see really different results? http://cassci.datastax.com/job/ptnapoleon-trunk-dtest/6/testReport/

smccarthy788:
Running new test fixes here: https://cassci.datastax.com/job/ptnapoleon-trunk-dtest/7/

Expecting bootstrap_test:TestBootstrap.local_quorum_bootstrap_test to fail. This is a genuinely broken test. Will file a follow up ticket on merge.

ptnapoleon:
+1

PR #511, opened 2016-06-10
Fix ccmlib example in README

penberg:
The Cluster class is now in 'ccmlib.cluster' so fix up the example
snippet in README accordingly.


ptnapoleon:
+1. thank you

PR #515, opened 2016-06-14
make log watching more amenable to resource cleanup

knifewine:
the main gist here, is now hand back a full-fledged thread to the caller who can .join() the thread when they are done watching logs (typically after cluster stop, but before file cleanup).

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -152,23 +152,23 @@ class Cluster(object):
 
                 return errordata
 
+            def scan_and_report(self):
+                errordata = self.scan()
+
+                if errordata:
+                    on_error_call(errordata)
+
             def run(self):
                 print_("Log-watching thread starting.")
 
                 # run until stop gets requested by .join()
                 while not self.req_stop_event.is_set():
-                    errordata = self.scan()
-
-                    if errordata:
-                        on_error_call(errordata)
-
+                    self.scan_and_report()
                     time.sleep(interval)
 
                 try:
                     # do a final scan to make sure we got to the very end of the files
-                    errordata = self.scan()
-                    if errordata:
-                        on_error_call(errordata)
+                    self.scan_and_report()
                 finally:
                     print_("Log-watching thread exiting.")
                     # done_event signals that the scan completed a final pass
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -132,8 +132,14 @@ class Cluster(object):
 
                 try:
                     for node in self.cluster.nodelist():
-                        errors = node.grep_log_for_errors_from(seek_start=self.log_positions[node.name])
-                        self.log_positions[node.name] = node.mark_log()
+                        scan_from_mark = self.log_positions[node.name]
+                        next_time_scan_from_mark = node.mark_log()
+                        if (next_time_scan_from_mark == scan_from_mark):
+                            # log hasn't advanced, nothing to do for this node
+                            continue
+                        else:
+                            errors = node.grep_log_for_errors_from(seek_start=scan_from_mark)
+                        self.log_positions[node.name] = next_time_scan_from_mark
                         if errors:
                             errordata[node.name] = errors
                 except IOError as e:
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -106,36 +106,78 @@ class Cluster(object):
 
     def actively_watch_logs_for_error(self, on_error_call, interval=1):
         """
-        Begins a timed daemon thread that scans logs for errors every interval seconds.
+        Begins a thread that scans logs for errors every interval seconds.
 
-        If an error is seen and the callback is executed, it will be called with an
-        OrderedDictionary mapping node name to a list of error lines.
+        If an error is seen, the callback is executed with an OrderedDictionary
+        mapping node name to a list of error lines.
+
+        Returns the thread itself, which should be .join()'ed to wrap up execution,
+        otherwise will run until the main thread exits.
         """
-        log_positions = defaultdict(int)
+        class LogWatchingThread(threading.Thread):
+            """
+            This class is embedded here for now, because it is used only from
+            within Cluster, and depends on cluster.nodelist().
+            """
+            def __init__(self, cluster):
+                super(LogWatchingThread, self).__init__()
+                self.cluster = cluster
+                self.daemon = True  # set so that thread will exit when main thread exits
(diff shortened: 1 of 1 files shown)


knifewine:
Running a trial of dtest/ccm changes here: http://cassci.datastax.com/view/Upgrades/job/upgrade_tests-all-custom_branch_runs/28/
(then should probably do a full dtest run, too)

knifewine:
dtest run here: http://cassci.datastax.com/view/Dev/view/knifewine/job/trunk_dtest_with_log_watching_cs502/1/

knifewine:
links above updated for newer jobs (upgrade failed mysteriously on first try, dtest had a little bug I fixed).

knifewine:
with #518 merged, running the upgrade trial job again here: http://cassci.datastax.com/view/Upgrades/job/upgrade_tests-all-custom_branch_runs/31/

knifewine:
and yet another one (job infra flaked out) http://cassci.datastax.com/view/Upgrades/job/upgrade_tests-all-custom_branch_runs/32/

knifewine:
@ptnapoleon ping

ptnapoleon:
Should I be scared that all these jobs fail?

ccmlib/cluster.py
```diff
@@ -106,36 +106,78 @@ def set_install_dir(self, install_dir=None, version=None, verbose=False):
 
     def actively_watch_logs_for_error(self, on_error_call, interval=1):
```
ptnapoleon: This only scans system.log yes? Can you make that explicit inside the docstring?

ccmlib/cluster.py
```diff
@@ -106,36 +106,78 @@ def set_install_dir(self, install_dir=None, version=None, verbose=False):
...
+                errordata = OrderedDict()
 
-        def grep_and_mark_all_logs():
-            errordata = OrderedDict()
+                try:
+                    for node in self.cluster.nodelist():
+                        errors = node.grep_log_for_errors_from(seek_start=self.log_positions[node.name])
+                        self.log_positions[node.name] = node.mark_log()
+                        if errors:
+                            errordata[node.name] = errors
```
ptnapoleon: Doesn't this risk overwriting previous errors?

knifewine:
> Should I be scared that all these jobs fail?
I don't think so, the big dtest run was just fine. I think openstack keeps flaking the upgrade jobs unrelatd to the code, but I will make sure we get a clean run there before merging.

ccmlib/cluster.py
ptnapoleon: I kinda see why not now, but I don't like the contract this is establishing.

Do we only care to do a thing the first time errors are found? That isn't really what the docstring or the `run` implementation implies.

ccmlib/cluster.py
```diff
@@ -106,36 +106,78 @@ def set_install_dir(self, install_dir=None, version=None, verbose=False):
...
+                    errordata = self.scan()
+
+                    if errordata:
+                        on_error_call(errordata)
+
+                    time.sleep(interval)
+
+                try:
+                    # do a final scan to make sure we got to the very end of the files
+                    errordata = self.scan()
```
ptnapoleon: Do we want to break out the this duped code into a tiny function? WWJD, where J == Jim, aka @mambocab

ccmlib/cluster.py
```diff
@@ -106,36 +106,78 @@ def set_install_dir(self, install_dir=None, version=None, verbose=False):
...
+                    if errordata:
+                        on_error_call(errordata)
+                finally:
+                    print_("Log-watching thread exiting.")
```
ptnapoleon: This is super not your fault, since I let this sit for 6 years before reviewing, but can you use the new logging code inside `common.py` instead of printing?

ccmlib/cluster.py
knifewine: pushed.

ccmlib/cluster.py
```diff
@@ -106,36 +106,78 @@ def set_install_dir(self, install_dir=None, version=None, verbose=False):
...
+
+            def scan(self):
+                errordata = OrderedDict()
 
-        def grep_and_mark_all_logs():
-            errordata = OrderedDict()
+                try:
+                    for node in self.cluster.nodelist():
+                        errors = node.grep_log_for_errors_from(seek_start=self.log_positions[node.name])
+                        self.log_positions[node.name] = node.mark_log()
```
knifewine: good catch. just pushed fix along the lines you proposed.

ccmlib/cluster.py
knifewine: no, the dict used is short-lived, and just represents errors for that pass. each node is checked once, and keyed by its name, so unless nodes somehow collide names I think this is safe.

ccmlib/cluster.py
knifewine: done.

ccmlib/cluster.py
knifewine: glad to \o/

knifewine: pushed.

ccmlib/cluster.py
knifewine: > Do we only care to do a thing the first time errors are found?

I think so. To be clear this is running as a loop in the thread, repeatedly. The callback will be called _each time_ new errors are discovered in a log chunk. It does not make multiple calls to report the same error repeatedly, by design.

> That isn't really what the docstring or the run implementation implies.

As far as I can tell the docstring is accurate, and not sure what you mean by what "the run implementation implies". I don't see the discrepancy -- can you be more specific about what you would like to change? (if you still want anything to change).

knifewine: pushed a more descriptive docstring.

ccmlib/cluster.py
```diff
@@ -106,36 +106,86 @@ def set_install_dir(self, install_dir=None, version=None, verbose=False):
...
-            errordata = OrderedDict()
+                try:
+                    for node in self.cluster.nodelist():
+                        scan_from_mark = self.log_positions[node.name]
+                        next_time_scan_from_mark = node.mark_log()
+                        if (next_time_scan_from_mark == scan_from_mark):
+                            # log hasn't advanced, nothing to do for this node
+                            continue
+                        else:
+                            errors = node.grep_log_for_errors_from(seek_start=scan_from_mark)
```
ptnapoleon: I am now confident that this logic only has a risk of overscanning, which is fine

PR #518, opened 2016-06-23
make error grep work on consecutive errors

knifewine:
add test for the same


ccmlib/node.py
```diff
@@ -1867,17 +1866,23 @@ def _get_load_from_info_output(info):
...
-            try:
-                it, peeker = itertools.tee(it)
-                while 'INFO' not in next(peeker):
-                    matchings[-1].append(next(it))
-            except StopIteration:
-                break
-    return matchings
+    def log_line_category(line):
+        match = re.search(r'(INFO|DEBUG|WARN|ERROR)', line)
+        return match.group(0) if match else 'ERROR_EXTRA'
```
mambocab: Love pulling this out into a method. I think `None` would be a better return value for the case where `not match`.

ccmlib/node.py
```diff
@@ -1867,17 +1866,23 @@ def _get_load_from_info_output(info):
...
+
+    matches = []
+
+    loglines = log.splitlines()
+
+    for line_num, line in enumerate(loglines):
+        if log_line_category(line) == 'ERROR':
+            matches.append([line])
+
+            for next_line_num in xrange(line_num+1, len(loglines)):
```
mambocab: Could you use `range` here instead of `xrange`, for Py3 compatibility?

mambocab:
This is great; the `enumerate` loop makes a lot more sense than the old `tee` thing. Go ahead and address my nits and merge.

ccmlib/node.py
knifewine: oh, good catch, forgot that's the new default behaviour anyway. will do.

ccmlib/node.py
knifewine: ok, so to determine if a line is an error continuation you want me to check for None rather than 'ERROR_EXTRA'? just want to make sure I got that right.

knifewine: pushed.

mambocab:
+1 to the changes, feel free to squash and merge

mambocab:
@ptnapoleon we don't have permissions to merge, can you do the honors please?

PR #519, opened 2016-06-23
Support specifying environment variables

thobbs:
As a start, this will be useful for specifying `CASSANDRA_LIBJEMALLOC` in
order to speed up dtests


ptnapoleon:
+1

PR #527, opened 2016-07-01
Fix issue when DSE nodes are removed/cleaned and ds agent is not installed

snazy:
Variable `pidfile` is defined, when the DS agent directory is evaluated.
However, if the agent is not installed, `pidfile` is not defined, which leads to an error.

Unsure though, whether it's an error that the agent was not installed (used DSE 5 + OpsC 6).


ptnapoleon:
+1

PR #529, opened 2016-07-11
Validate node name on cmdline node add so that we don't mess up the state on node initialization

jkni:
When we add a node on the command line, if we duplicate the name of an existing node, it blows away state on disk in the node constructor but then fails to add the node to the cluster when the cluster validates the name.

There isn't a clear way I can see to disentangle this sequence without reworking the whole cluster/node abstraction. For now, we can duplicate the name check so that fat fingering a node name doesn't mess things up too badly.

### DIFF
--- a/ccmlib/cmds/cluster_cmds.py
+++ b/ccmlib/cmds/cluster_cmds.py
@@ -254,6 +254,11 @@ class ClusterAddCmd(Cmd):
             parser.print_help()
             exit(1)
 
+        if self.name in self.cluster.nodes:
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Do we also need/want to check if a node is added via the API?

jkni:
Do you mean in the node constructor or the cluster add? We check in the cluster add, but not in the node constructor. If we do it in the node constructor, we could remove this check from the commandline, but I'm not sure if there are time we call the node constructor and want to overwrite state on disk.

ptnapoleon:
Ehh, this is sufficient then for now. Thanks!

PR #530, opened 2016-07-11
use clock for timeout, not a counter

mambocab:
I've run this against the code from [this pr](https://github.com/riptano/cassandra-dtest/pull/1086), and it seems to work.


ptnapoleon:
+1

PR #532, opened 2016-07-15
Allow 3rd party extensions to be installed

mike-tr-adamson:


ptnapoleon:
I'm +1, but running dtest with these changes to double check everything is good. I'm going to bump the version once this is merged.
https://cassci.datastax.com/view/Dev/view/ptnapoleon/job/ptnapoleon-trunk-dtest/21/

PR #534, opened 2016-07-31
Updated DSE environment setting for 5.0 and above

mike-tr-adamson:
Allow control of the `SPARK_WORKER` resources and a specific 5.0 and above change for the introduction of HADOOP 2 support.


ptnapoleon:
Sorry for the delay, I was OOO. +1!

PR #535, opened 2016-07-31
Changes to DSE node configuration for 5.0

mike-tr-adamson:
These include:
  hadoop2_client configuration
  correct configuration of the binary folders
  setting DSE_HOME in the dse-env.sh (used by a number of CLI tools)
  settings for Spark 1.6 integration


ptnapoleon:
Sorry for the delay, I was OOO. +1!

PR #538, opened 2016-08-08
remove nodetool call from versionfrombuild cmd

mambocab:
This removes all `nodetool` logic. It wasn't necessary in the first place and makes the command fragile.


ptnapoleon:
+1

PR #541, opened 2016-08-26
add grep_nodes_for_patterns method

smccarthy788:
@ptnapoleon & @mambocab, how much of a need do you feel there is for this function?

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -16,6 +16,7 @@ import time
 import logging
 
 import yaml
+from distutils.version import LooseVersion
 from six import print_
 
 BIN_DIR = "bin"
@@ -74,6 +75,34 @@ class ArgumentError(CCMError):
 
 class UnavailableSocketError(CCMError):
     pass
+
+
+class LogPatternToVersion(object):
+
+    def __init__(self, versions_to_patterns, default_pattern=None):
+        self.versions_to_patterns, self.default_pattern = versions_to_patterns, default_pattern
+
+    def __call__(self, version):
+        keys_less_than_version = [k for k in self.versions_to_patterns if k <= version]
+
+        if not keys_less_than_version:
+            if self.default_pattern is not None:
+                return self.default_pattern
+            else:
+                raise ValueError("Some kind of default pattern must be specified!")
+
+        return self.versions_to_patterns[max(keys_less_than_version, key=lambda v: LooseVersion(v) if not isinstance(v, LooseVersion) else v)]
+
+    def __repr__(self):
+        return str(self.__class__) + "(versions_to_patterns={}, default_pattern={})".format(self.versions_to_patterns, self.default_pattern)
+
+    @property
+    def patterns(self):
+        return list(self.versions_to_patterns.values())
+
+    @property
+    def versions(self):
+        return list(self.versions_to_patterns)
 
 
 def get_default_path():
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -98,7 +98,10 @@ class LogPatternToVersion(object):
 
     @property
     def patterns(self):
-        return list(self.versions_to_patterns.values())
+        patterns = list(self.versions_to_patterns.values())
+        if self.default_pattern:
+            patterns = patterns + [self.default_pattern]
+        return patterns
 
     @property
     def versions(self):
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -8,7 +8,7 @@ import subprocess
 import sys
 import threading
 import time
-from collections import OrderedDict, defaultdict
+from collections import OrderedDict, defaultdict, namedtuple
 
 import yaml
 from six import iteritems, print_
@@ -657,3 +657,29 @@ class Cluster(object):
     def enable_pwd_auth(self):
         self._config_options['authenticator'] = 'PasswordAuthenticator'
         self._update_config()
+
+    def timed_grep_nodes_for_patterns(self, versions_to_patterns, timeout, filename="system.log"):
+        """
+        Searches all nodes in the cluster for a specific regular expression based on the node's version.
+        Params:
+        @versions_to_patterns : an instance of LogPatternToVersionMap, specifying the different log patterns based on a node's version.
+        @version : the earliest version the new pattern was introduced.
+        @timeout : the amount of time to spend searching the logs for.
+        @filename : the name of the file to search for the patterns. Defaults to "system.log".
+
+        Returns the first node where the pattern was found, along with the matching lines.
+        Raises a TimeoutError if the pattern is not found within the specified timeout period.
+        """
+        end_time = time.time() + timeout
+        while True:
+            if time.time() > end_time:
+                raise TimeoutError(time.strftime("%d %b %Y %H:%M:%S", time.gmtime()) +
+                                   " Unable to find: " + versions_to_patterns.patterns + " in any node log within " + str(timeout) + "s")
+
+            for node in self.nodelist():
+                pattern = versions_to_patterns(node.get_cassandra_version())
+                matchings = node.grep_log(pattern, filename)
+                if matchings:
+                    ret = namedtuple('Node_Log_Matching', 'node matchings')
+                    return ret(node=node, matchings=matchings)
+            time.sleep(1)


mambocab:
This API's a little clunky IMO -- as currently PR'd, I think I prefer writing one-off functions as we need to do this. That said, I'd be down for one with a really great API.

mambocab:
My answer isn't definitive, others' opinions totally welcome ofc.

ptnapoleon:
I'm indifferent

smccarthy788:
I agree with Jim, the API is gross. Ideally, I think it'd be nice to specify any number of versions, and any number of patterns. We'd need some kind of map {version -> pattern}. Maybe if the map is signifying which version is the min version for a particular pattern. Then we'd just have some sort of `map.get_version()` method. I think all of this would best be implemented in a class, which asks the question of where: in `cluster.py` or `common.py`? Those seem the most obvious to me.

I'll turn this ramble into code so you can see what it might look like.

smccarthy788:
Something like this: https://gist.github.com/smccarthy788/09610b2552f7619d9f098aa662d3080b

mambocab:
I like the idea of representing the mapping as some kind of data structure. If you wanna write a dict with custom behavior, the way to do it, I think, is by subclassing the `MutableMapping` abstract base class:

https://docs.python.org/2/library/collections.html#collections.MutableMapping

Should make it easy to write an object that gives you the right value when you ask something like `pattern_for_version_map['3.0.6']`.

This is a cool idea - I'd want to see this in use in a couple call sites before +1ing it though.

mambocab:
Another option, one that might be a little more readable than the `dict`like implementation, would be implementing a little "function that returns a function with closed-over arguments" sort of interface with a simple class with a `__call__` method. [`get_rate_limited_function`](https://github.com/riptano/cassandra-dtest/blob/0a45bada3095bbb1e7c82f22224010286def74da/tools/funcutils.py#L4-L29) is an example of one of these classes.

smccarthy788:
@mambocab, I implemented the `MutableMapping` before your comment. You can see the implementation in this PR, and an example call site here: https://github.com/riptano/cassandra-dtest/commit/d6ac93befa4669a8eb7516e32a46d4a64c7217de

smccarthy788:
@mambocab, I've implemented this in a more closure-esque style. I've kept the `get_versions()` and `get_patterns()` methods. I think they could be useful.

ccmlib/common.py
```diff
@@ -76,6 +77,36 @@ class UnavailableSocketError(CCMError):
...
+        return pattern
+
+    def __iter__(self):
+        return iter(self.versions_to_patterns)
+
+    def __len__(self):
+        return len(self.versions_to_patterns)
+
+    def __repr__(self):
+        return str(self.versions_to_patterns)
```
mambocab: I think I'd rather have this look more like `LogPatternToVersion(versions_to_patterns=...)`. You should get the class name from `self.__class__` rather than hardcoding it.

ccmlib/common.py
```diff
@@ -76,6 +77,36 @@ class UnavailableSocketError(CCMError):
...
 
 
+class LogPatternToVersion(object):
+    def __init__(self, patterns_to_versions):
+        assert type(patterns_to_versions) is dict, "LogPatternToVersion takes a dictionary."
+        self.versions_to_patterns = patterns_to_versions
+
+    def __call__(self, version):
+        pattern = None
+        for key in self.versions_to_patterns.keys():
```
mambocab: I don't think this will work correctly, as you can't depend on iteration order to be anything in particular.

ccmlib/common.py
```diff
@@ -76,6 +77,36 @@ class UnavailableSocketError(CCMError):
     pass
 
 
+class LogPatternToVersion(object):
+    def __init__(self, patterns_to_versions):
+        assert type(patterns_to_versions) is dict, "LogPatternToVersion takes a dictionary."
+        self.versions_to_patterns = patterns_to_versions
```
mambocab: Why the name change here? I think the argument name should be `version_to_patterns` here.

ccmlib/common.py
```diff
@@ -76,6 +77,36 @@ class UnavailableSocketError(CCMError):
     pass
 
 
+class LogPatternToVersion(object):
+    def __init__(self, patterns_to_versions):
+        assert type(patterns_to_versions) is dict, "LogPatternToVersion takes a dictionary."
```
mambocab: You would do this with a `isinstance` call, but I don't think this is a use case to do type checking for. Just let it blow up if the passed-in thing doesn't work.

ccmlib/common.py
```diff
@@ -76,6 +77,36 @@ class UnavailableSocketError(CCMError):
...
+
+    def __len__(self):
+        return len(self.versions_to_patterns)
+
+    def __repr__(self):
+        return str(self.versions_to_patterns)
+
+    @property
+    def patterns(self):
+        return [self.versions_to_patterns[version] for version in self.__iter__()]
```
mambocab: I think this could be simplified to `return list(self.version_to_patterns.values())`[

ccmlib/common.py
```diff
@@ -76,6 +77,36 @@ class UnavailableSocketError(CCMError):
...
+
+    @property
+    def versions(self):
+        return [version for version in self.__iter__()]
```
mambocab: And this (I think) could actually be simplified to `list(self.versions_to_patterns)` -- iterating over a dictionary gives you the keys.

ccmlib/common.py
smccarthy788: `self.__class__` returns ```<class '__main__.LogPatternToVersion'>```

Not sure if this is what you want

ccmlib/common.py
```diff
@@ -76,6 +77,36 @@ class UnavailableSocketError(CCMError):
...
+class LogPatternToVersion(object):
+
+    def __init__(self, versions_to_patterns):
+        self.versions_to_patterns = versions_to_patterns
+
+    def __call__(self, version):
+        pattern = None
+        for key in sorted(self.versions):
+                if version >= key:
+                    pattern = self.versions_to_patterns[key]
```
mambocab: Still not sure about this loop, sorry -- it's just kinda hard to reason about. I think a more readable strategy here would be something like

```python
greatest_key_less_than_version = max(k for k in self.versions_to_patterns if k < version)
return self.versions_to_patterns[greatest_key_less_than_version]
```

The other thing to be concerned about is that we should somehow make sure that we compare `LooseVersion`s here. What API do we want? I'm tempted to have logic like

`LooseVersion(v) if not isinstance(v, LooseVersion) else v`. @ptnapoleon, thoughts?

ptnapoleon: I like that logic

ccmlib/common.py
```diff
@@ -76,6 +77,36 @@ class UnavailableSocketError(CCMError):
     pass
 
 
+class LogPatternToVersion(object):
+
+    def __init__(self, versions_to_patterns, default_pattern):
+        self.versions_to_patterns = versions_to_patterns
+        if "0" not in self.versions_to_patterns:
```
smccarthy788: Here I'm working on the assumption someone might add a version "0" pattern, though it seems... unlikely.

ccmlib/common.py
```diff
@@ -76,6 +77,36 @@ class UnavailableSocketError(CCMError):
...
+    def __repr__(self):
+        return str(self.__class__) + "(versions_to_patterns={})".format(self.versions_to_patterns)
+
+    @property
+    def patterns(self):
+        return list(self.versions_to_patterns.values())
+
+    @property
+    def versions(self):
+        return list(self.versions_to_patterns)
```
smccarthy788: Would this be better if it returned a list of `LooseVersion`s? @mambocab?

ccmlib/common.py
mambocab: This is clever... a little too clever :P The problem here is that the fallback becomes emergent behavior, rather than explicit behavior. If we were writing some kind of high-performance code, that'd be worth the loss in readability, but I don't think it is here. I'd add a `default` attribute to the object itself and use that throughout. the `__repr__` should account for that.

And sorry to waffle on this, but I think that `default` should be an optional kwarg. I don't really like requiring callers to specify both a dict and a fallback, even in cases when they know they won't need the fallback. In cases where `__call__` would fall back and `default` wasn't specified, `__call__` should blow up.

At this point, I think that `__iter__` and `__len__` aren't particularly meaningful for this object, since it's not really a simple map, and I'm not sure what I'd use `patterns` and `versions` for. As to whether `versions` should return a list of `LooseVersions` or not -- I think the thing we should do is wait to see if we need it and then write the method up to that need.

smccarthy788: I think I need some clarification. 

The default pattern is a base case for our search through `versions_to_patterns`, i.e. when we can't find any version key less than the specified version, we need some answer. We need this because `max(...)` will blow up if we pass it an empty list as its iterable. In this case, if the default pattern isn't a part of `versions_to_patterns`, we will generate an empty list and blow up `max(...)`. If, in this same case, default is a property of the object, we have to catch the error `max(...)` is throwing and check if default exists.

I think that logic looks [something like this.](https://gist.github.com/smccarthy788/ece0ab0f4b9aa42ebb916c6a1ce5719a)

However, this implementation would preclude the default pattern from being `None`, which I think is the most correct answer in situations where a pattern didn't exist before some version. I suspect this is just an implementation error though. (actually, by the time I've posted this I think I've found a way to fix it, but it blows up `__repr__`)

I agree with you on `versions`, `__iter__`, and `__len__`, however, I use `patterns` in the `TimeoutError` implemented in `cluster.py`. I think `x.patterns` is more readable than `x.versions_to_patterns.values()`

mambocab: I'm not sure I follow how the existing implementation precludes `default` being `None`, but I think you're on the right track. I think something like

```python
def __init__(self, versions_to_patterns, default=None):
    self.version_to_patterns, self.default = versions_to_patterns, default

def __call__(self, version):
    keys_less_than_version = [k for k in self.versions_to_patterns if k <= version]
    
    if not keys_less_than_version:
        if self.default is not None:
            return 
        raise ValueError(...)
    return self.versions_to_patterns[max(keys_less_than_version, key=...)
```

would do the trick and be readable.

smccarthy788: I see it. I just now realized we also need to compare `LooseVersion`s when reducing the keys for `keys_less_than_version`.

mambocab: Whoop, yeah, good catch!

smccarthy788:
@mambocab, I've refactored as to your suggestions.

ccmlib/common.py
mambocab: No, I think it's fine as-is.

ccmlib/common.py
```diff
@@ -76,6 +77,34 @@ class UnavailableSocketError(CCMError):
...
+                raise ValueError("Some kind of default pattern must be specified!")
+
+        return self.versions_to_patterns[max(keys_less_than_version, key=lambda v: LooseVersion(v) if not isinstance(v, LooseVersion) else v)]
+
+    def __repr__(self):
+        return str(self.__class__) + "(versions_to_patterns={}, default_pattern={})".format(self.versions_to_patterns, self.default_pattern)
+
+    @property
+    def patterns(self):
+        return list(self.versions_to_patterns.values())
```
mambocab: This should add the default value if there is one.

mambocab:
Sweet, thank you. Code LGTM; can we get a merge from @ptnapoleon?

PR #542, opened 2016-08-31
implement wait_for_thrift_interface

smccarthy788:
@ptnapoleon & @mambocab, please review.

There is one thing I'm not sure about, which I had just copied over from `wait_for_binary_interface()`: the conditional based on version number 1.2 in line 508. Is this the correct version?

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -497,6 +497,17 @@ class Node(object):
             warnings.warn("Binary interface %s:%s is not listening after 10 seconds, node may have failed to start."
                           % (binary_itf[0], binary_itf[1]))
 
+    def wait_for_thrift_interface(self, **kwargs):
+        """
+        Waits for the Thrift interface to be listening.
+
+        Emits a warning if not listening after 10 seconds.
+        """
+
+        thrift_itf = self.network_interfaces['thrift']
+        if not common.check_socket_listening(thrift_itf, timeout=10):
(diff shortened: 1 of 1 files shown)


ccmlib/node.py
```diff
@@ -497,6 +497,21 @@ def wait_for_binary_interface(self, **kwargs):
             warnings.warn("Binary interface %s:%s is not listening after 10 seconds, node may have failed to start."
                           % (binary_itf[0], binary_itf[1]))
 
+    def wait_for_thrift_interface(self, **kwargs):
+        """
+        Waits for the Thrift interface to be listening.  If > 1.2 will check
+        log for 'Starting listening for CQL clients' before checking for the
```
ptnapoleon: This comment is now unrelated. Just keep the first sentence of this paragraph.

ccmlib/node.py
```diff
@@ -497,6 +497,21 @@ def wait_for_binary_interface(self, **kwargs):
...
+        interface to be listening.
+
+        Emits a warning if not listening after 10 seconds.
+        """
+        if self.cluster.version() >= '1.2':
```
ptnapoleon: Drop the `if` check

smccarthy788:
@ptnapoleon, I have addressed your review comments.

ccmlib/node.py
ptnapoleon: My bad, I meant only drop the if check, not the code inside the if check.

ptnapoleon:
+1

PR #547, opened 2016-09-13
allow for an empty string to be represented as the default pattern

smccarthy788:
@ptnapoleon, @knifewine: I think this should address the issue raised in https://github.com/riptano/cassandra-dtest/pull/1311#issuecomment-246747914

We should be able to get the default pattern if it is falsy, but exists.


knifewine:
+1

PR #548, opened 2016-09-14
Loose versions

ptnapoleon:
@mambocab to review

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -237,7 +237,7 @@ class Node(object):
 
     def get_base_cassandra_version(self):
         version = self.get_cassandra_version()
-        return float(version[:version.index('.') + 2])
+        return float('.'.join(version.vstring.split('.')[:2]))
 
     def set_configuration_options(self, values=None):
         """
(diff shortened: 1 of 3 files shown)


ccmlib/node.py
```diff
@@ -237,7 +236,7 @@ def get_cassandra_version(self):
 
     def get_base_cassandra_version(self):
         version = self.get_cassandra_version()
-        return float(version[:version.index('.') + 2])
+        return float(version.vstring[:version.vstring.index('.') + 2])
```
mambocab: This seems to have never worked correctly for `W.XY.Z`:

```python
In [5]: for version in ('3.1', '3.0.1', '3.10.1'):
   ...:     print version, '\t', version[:version.index('.') + 2]
   ...:     
3.1 	3.1
3.0.1 	3.0
3.10.1 	3.1
```

`'.'.join(version.vstring.split('.')[:2])` should do the trick:

```python
In [8]: for version in ('3.1', '3.0.1', '3.10.1'):
   ...:     print version, '\t', '.'.join(version.split('.')[:2])
   ...:     
3.1 	3.1
3.0.1 	3.0
3.10.1 	3.10
```

I don't like that this returns a float, but that's out of scope I think.

ptnapoleon:
Okay @mambocab, ready for re-review

PR #549, opened 2016-09-14
allow for verbose cli. pep8 clean

smccarthy788:
@ptnapoleon, I found a little bug. Trying to run `ccm node1 cli` would result in an argument error, since the `run_cli` method doesn't take the `verbose` argument anymore.

I'm happy to bikeshed over how this output is printed.

### DIFF
--- a/ccmlib/cmds/node_cmds.py
+++ b/ccmlib/cmds/node_cmds.py
@@ -55,6 +55,7 @@ NODE_CMDS = [
     "byteman"
 ]
 
+
 def node_cmds():
     return NODE_CMDS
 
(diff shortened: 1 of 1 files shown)


smccarthy788:
After bikeshedding with myself for a minute, I find I like this output. It looks something like:

```
12:02 PM:~/cstar/ccm[arg-fix*]$ ccm node1 cli -v -x aslkjfashd
CLI OUTPUT:
-------------------------------
Connected to: "supcol" on 127.0.0.1/9160
Welcome to Cassandra CLI version 1.2.19

Type 'help;' or '?' for help.
Type 'quit;' or 'exit;' to quit.

[default@unknown] aslkjfashd;
[default@unknown] quit;

-------------------------------
CLI ERROR:
-------------------------------
Syntax error at position 0: no viable alternative at input 'aslkjfashd'
```

ccmlib/cmds/node_cmds.py
```diff
@@ -370,7 +371,12 @@ def validate(self, parser, options, args):
         self.cli_options = parser.get_ignored() + args[1:]
 
     def run(self):
-        self.node.run_cli(self.options.cmds, self.options.verbose, self.cli_options)
+        out, err, rc = self.node.run_cli(self.options.cmds, self.cli_options)
+        if self.options.verbose:
+            print "CLI OUTPUT:\n-------------------------------"
+            print out
+            print "-------------------------------\nCLI ERROR:\n-------------------------------"
+            print err
```
ptnapoleon: Use `print_` from `six` here, instead of just `print`

PR #554, opened 2016-09-20
make local slugs better

knifewine:
This proposes to remove the old form of local:somebranch (requiring LOCAL_GIT_REPO defined in env) and replacing it with a form of local:/some/path/:branch_name . This is to support more easily working with multiple local versions across the filesystem (in multiple git repos). The old way made it clumsy to do that.

I tested this by hand locally and it appears to function well from the command line. If there's any specific cases you can think of I'm happy to test more.

I think the changes here have a low risk outside of local: usage.

### DIFF
--- a/ccmlib/repository.py
+++ b/ccmlib/repository.py
@@ -47,7 +47,7 @@ def setup(version, verbose=False):
         version_dir = version_directory(version)
 
         if version_dir is None:
-            raise CCMError("Path provided in local slug appears invalid.")
+            raise CCMError("Path provided in local slug appears invalid ({})".format(path))
         return (version_dir, None)
 
     elif version.startswith('binary:'):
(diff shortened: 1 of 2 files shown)


ccmlib/repository.py
```diff
@@ -37,12 +36,19 @@ def setup(version, verbose=False):
...
+        try:
+            _, path, branch = version.split(':')
+        except ValueError:
+            raise CCMError("local version ({}) appears to be invalid. Please format as local:/some/path/:somebranch".format(version))
+
+        clone_development(path, version, verbose=verbose)
+        version_dir = version_directory(version)
+
+        if version_dir is None:
+            raise CCMError("Path provided in local slug appears invalid.")
```
mambocab: This could be a little more verbose -- maybe include the path and a little insight about how it's invalid.

ptnapoleon: I'm fine with no insight, that sounds infeasible to implement, but including the path would be great?

ptnapoleon:
Can you make Jim's change, and also bump the minor version number (in setup.py)? Thanks

ccmlib/repository.py
knifewine: added the path to that message so users can see what value is deemed bad.

knifewine:
Changes made -- could you guys take one more quick look? Philip, merge if it looks good to you!

PR #555, opened 2016-09-21
make cli args consistent through cluster and node. Closes #553

### LINKED ISSUE #553: Parameter mismatch between cluster.run_cli and node.run_cli
I think there might be a bug in the run_cli function for the cluster object.

It's calling the node's run_cli method here.

https://github.com/pcmanus/ccm/blob/master/ccmlib/cluster.py#L505
With 4 params (including self).
However node.run_cli only has 3 params including self.

https://github.com/pcmanus/ccm/blob/master/ccmlib/node.py#L826

This is going lead to a stack trace like this.

File "/usr/local/lib/python2.7/site-packages/ccmlib/cluster.py", line 504, in run_cli
    livenodes[0].run_cli(cmds, show_output, cli_options)
TypeError: run_cli() takes at most 3 arguments (4 given)

I imagine we should extend the node's run_cli method to utilize the new parameter.

smccarthy788:
@ptnapoleon, for your review. Are you seeing any other errors using `ccm cli` from the command line?

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -502,7 +502,7 @@ class Cluster(object):
         livenodes = [node for node in list(self.nodes.values()) if node.is_live()]
         if len(livenodes) == 0:
             raise common.ArgumentError("No live node")
-        livenodes[0].run_cli(cmds, show_output, cli_options)
+        return livenodes[0].run_cli(cmds, cli_options)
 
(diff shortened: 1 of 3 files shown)


ptnapoleon:
Lgtm, @GregBestland ?

GregBestland:
+9001 LGTM

PR #557, opened 2016-10-03
Use SpooledTemporaryFile for stderr in subprocess.

tolbertam:
DSE workloads can produce a lot of output on start.  In particular, it will print the full command line used.  With DSE, this can exceed 40,000 characters.

Using subprocess.PIPE has limitations, particularly that after 65535
characters, it may block the subprocess (see:
https://thraxil.org/users/anders/posts/2008/03/13/Subprocess-Hanging-PIPE-is-your-enemy/).

Using a temporary file works around this possible occurrence.

Also added a small change to prevent printing stderr multiple times.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -11,6 +11,7 @@ import stat
 import subprocess
 import sys
 import time
+import tempfile
 import warnings
 from collections import namedtuple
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Lgtm

PR #559, opened 2016-10-06
Actually clear data when node.clear(only_data) is called

ptnapoleon:
@mambocab or @smccarthy788  to review.

It's a breaking API change, but we use it literally nowhere. Also neither do the python driver tests. And anyone trying to use it would have noticed it does nothing.


smccarthy788:
Lgtm +1

PR #560, opened 2016-10-06
add version_slug to cluster config

mambocab:
When the cached checkout for a cluster is deleted, the cluster gets stranded -- you can't load it up with `ClusterFactory.load` or even `Cluster.remove` it, because it needs a cached repo backing it to do so. This adds a version_slug value to the cluster config that allows CCM to recreate the cached repo when it needs to. The following code fails before this commit and succeeds after it:

```
cluster = Cluster(
    path=DEFAULT_PATH,
    name=CLUSTER_NAME,
    install_dir=os.path.join(DEFAULT_PATH, CLUSTER_NAME),
    version='git:trunk',
    create_directory=True,
    verbose=True,
)

common.invalidate_cache()

loaded_cluster = ClusterFactory.load(
    DEFAULT_PATH, CLUSTER_NAME
) # fails before this commit, passes after
```

or, from the command line, 

```
( ( set -x ; ccm create repro-cs743 -v git:trunk && ccm invalidatecache && ccm remove ) ; rm -rf ~/.ccm )
```

(The final `rm` isn't necessary for repro, but remove the ccm cache, which will be messed up if you run it before this commit.)

I want to add tests for this somewhere, and I'll do that tomorrow or early next week if this general approach looks good. @ptnapoleon?

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -32,6 +32,7 @@ class Cluster(object):
         self.__log_level = "INFO"
         self.__path = path
         self.__version = None
+        self._version_slug = version
         self.use_vnodes = False
         # Classes that are to follow the respective logging level
         self._debug = []
@@ -588,6 +589,7 @@ class Cluster(object):
         config_map = {
             'name': self.name,
             'nodes': node_list,
+            'version_slug': self._version_slug,
             'seeds': seed_list,
             'partitioner': self.partitioner,
             'install_dir': self.__install_dir,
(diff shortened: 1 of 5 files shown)


ptnapoleon:
This approach looks good, some nits discussed offline

mambocab:
Running dtests using this here:

http://cassci.datastax.com/view/Dev/view/mambocab/job/mambocab-trunk-copy-dtest/2/

mambocab:
I'm guessing that ran into some OpenStack problems? Rerunning here:

http://cassci.datastax.com/view/Dev/view/mambocab/job/mambocab-trunk-copy-dtest/3/

mambocab:
I won't have time to continue working on this, and I think it's superceded by #562, so I'll close for now.

PR #561, opened 2016-10-10
Default to starting clusters with binary_proto=True

ptnapoleon:
Leaving node as false for now, since sometimes when we
start nodes, in dtest, we're doing so to expect them to fail.
But clusters should generally always work on startup. The case
where one wants an entire cluster to fail to start is an odd one

@mambocab to review

CI is generally good, so I just need a few changes to dtest for this to work

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -361,7 +361,7 @@ class Cluster(object):
             else:
                 node.show(only_status=True)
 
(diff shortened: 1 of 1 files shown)


mambocab:
+1

mambocab:
To elaborate: I had a look at how `wait_for_binary_proto` works and it looks like this change should just plug in nicely

PR #562, opened 2016-10-18
allow for user configured aliases

smccarthy788:
@mambocab & @ptnapoleon: please review.

Just a little tweak. I'm trying to allow users to define aliases of git urls in a configuration file. Seems easier than `-v git:smccarthy788/cassandra-fork` every time.

Config files should look like

```
alias_name = some.git.url
```

be located at `.ccm/config`, and can be used as `ccm create -v alias:alias_name ...`

### DIFF
--- a/ccmlib/repository.py
+++ b/ccmlib/repository.py
@@ -134,7 +134,7 @@ def clone_development(git_repo, version, verbose=False, alias=False):
         git_repo_name = 'local_{}'.format(git_repo)  # add git repo location to distinguish cache location for differing repos
         git_branch = version.split(':')[-1]  # last token on 'local:...' slugs should always be branch name
     elif alias:
-        git_repo_name = 'alias_{}'.format(version.split(':')[0])
+        git_repo_name = 'alias_{}'.format(version.split('/')[0].split(':')[-1])
         git_branch = version.split('/')[-1]
     else:
         git_repo_name = 'apache'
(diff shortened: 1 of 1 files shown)


ptnapoleon:
At `~/.ccm/config`, you mean?

ptnapoleon:
Maybe we should make it a .ini formatted file, with an `[alias]` section, for future proofing? There should be a python library that handles that

ptnapoleon:
I think we can drop the explicit need for `alias`? When we receive a slug that we don't know how to handle, we can just check the user specified aliases, rather than making the user explicitly mark it as an alias. You can just make it an `else` check instead of an `elif`. You'll have to move the `version in 'stable', 'oldstable` block higher up, though

smccarthy788:
@ptnapoleon, we could not be explicit about `alias`, but since so many other parts of ccm touch `repository.py` we either end up printing something like `alias not found in config file` everytime we run `ccm status | start | remove | etc.`, or we increase scope, and refactor `repository.py`.

The problem is we need something to specify we're trying to use `alias`. When we call `setup()` to do something other than actually setup the cluster, it counts on the fact that the `version` passed doesn't start with any of `git: local: binary:` etc., but that's also what you're trying to do by not specifying `alias:`, so `setup()` falls into just returning the cluster directory and version name.

It's easy enough to always check the config file for the alias, but if the alias isn't there (as is the case when running `start()` ), I have to catch `ConfigParser.NoOptionError`. Since we can't determine if we're trying to actually setup a cluster or just get the cluster directory, handling the error is just to print a warning. If we're trying to setup a cluster and we don't find the alias, it will end up just falling into the default 'download cassandra from git', which is also not what we want.

ptnapoleon:
Sounds like we're using `alias:` explicitly until we refactor this, then

ptnapoleon:
This lgtm. @mambocab ?

mambocab:
Smoke tested it locally, works for me, doesn't break the `git:` slug. +1.

PR #566, opened 2016-12-13
Don't include removed thrift options when running against C* 4.0

pcmanus:
This is to support https://issues.apache.org/jira/browse/CASSANDRA-11115, and work at least as far as cassandra-dtests are involved on 4.0. Hopefully this doesn't break anything for earlier versions.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -136,7 +136,10 @@ class Node(object):
             binary_interface = None
             if 'binary' in itf and itf['binary'] is not None:
                 binary_interface = tuple(itf['binary'])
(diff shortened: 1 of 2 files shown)


ccmlib/node.py
```diff
@@ -514,6 +522,9 @@ def wait_for_thrift_interface(self, **kwargs):
 
         Emits a warning if not listening after 10 seconds.
         """
+        if self.cluster.version() < '4':
```
ptnapoleon: Shouldn't this be `>=` 4?

pcmanus: Of course, will fix, thanks.

ptnapoleon:
+1 after that small fix

PR #567, opened 2016-12-16
Copy wait_for_any_log from dtest to ccm

djbowers:
Copy wait_for_any_log from dtest to ccm.common (verbatim, without unused reference to self) and as a convenience method on ccm.cluster. Also import TimeoutError from ccmlib.node into ccm.common.


ccmlib/common.py
```diff
@@ -700,3 +702,27 @@ def is_intlike(obj):
...
+    """
+    for _ in range(timeout):
+        for node in nodes:
+            found = node.grep_log(pattern, filename=filename)
+            if found:
+                return node
+        time.sleep(1)
+
+    raise TimeoutError(time.strftime("%d %b %Y %H:%M:%S", time.gmtime()) +
+                       " Unable to find: " + repr(pattern) + " in any node log within " + str(timeout) + "s")
```
ptnapoleon: There should be a new line at the end of this file.

ccmlib/cluster.py
```diff
@@ -688,3 +688,26 @@ def timed_grep_nodes_for_patterns(self, versions_to_patterns, timeout_seconds, f
...
+        """
+        for _ in range(timeout):
+            for node in nodes:
+                found = node.grep_log(pattern, filename=filename)
+                if found:
+                    return node
+            time.sleep(1)
+
+        raise TimeoutError(time.strftime("%d %b %Y %H:%M:%S", time.gmtime()) +
+                           " Unable to find: " + repr(pattern) + " in any node log within " + str(timeout) + "s")
```
ptnapoleon: There should also be a new line at the end of this file.

ccmlib/cluster.py
```diff
@@ -688,3 +688,26 @@ def timed_grep_nodes_for_patterns(self, versions_to_patterns, timeout_seconds, f
                     ret = namedtuple('Node_Log_Matching', 'node matchings')
                     return ret(node=node, matchings=matchings)
             time.sleep(1)
+
+    def wait_for_any_log(self, nodes, pattern, timeout, filename='system.log'):
```
ptnapoleon: What you'll want to do here, rather than essentially duplicating the function, is to wrap the common.py function. Additionally, you don't want to take a list of nodes, but just use `cluster.nodelist()`.

So, in pseudocode, it would roughly look like:
```
def wait_for_any_log(self, pattern, timeout, filename='system.log'):
   return common.wait_for_any_log(self.nodelist(), pattern, timeout, filename=filename
```

ptnapoleon:
Looks great!

PR #570, opened 2016-12-21
Cleanup windows gating

djbowers:


ccmlib/common.py
```diff
@@ -328,6 +328,13 @@ def is_win():
     return sys.platform in ("cygwin", "win32")
 
 
+def is_modern_windows_install(version):
```
ptnapoleon: So, I'm a little worried about type handling here. Some of the callsites we replaced were using strings, and others used ints. Could you add unit tests for this?

ptnapoleon: Could you add a docstring to this function, explaining what it is? Can be as brief as:
```
"""
The 2.1 release line was when Cassandra received beta windows support. Many features are gated based on that added compatibility.
"""
```

tests/test_common.py
```diff
@@ -22,5 +25,16 @@ def test_normalize_interface(self):
         normalized = common.normalize_interface(('fd6d:404d:54cb::1', 9042))
         self.assertEqual(normalized, ('fd6d:404d:54cb:0:0:0:0:1', 9042))
 
+    @patch('ccmlib.common.is_win')
+    def test_is_modern_windows_install(self, mock_is_win):
```
ptnapoleon: Could you add another set of assertions, for a version that is higher than 2.1? 3.12 or 3.0.9 should do

Issue #556, opened 2016-09-26
Error downloading version: 'TarFile' object is not an iterator


Shaac:
Hello,

Since commit `f95de48 - Resolve all py3 compat warnings`, integrated in release 2.3.3, I have a `TypeError` when downloading and extracting cassandra: `WARNING Downloading 2.2.1 failed, due to 'TarFile' object is not an iterator. Trying to build from git instead.`

I ran the function `repository.download_version('2.2.1')` locally to get a full traceback:

    14:34:26,877 ccm INFO Extracting /tmp/ccm-kzjt40s_.tar.gz as version 2.2.1 ...
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
      File "/home/user/ccm/ccmlib/repository.py", line 278, in download_version
        dir = next(tar).name.split("/")[0]
    TypeError: 'TarFile' object is not an iterator

It seems that `next` can not be called on a `TarFile` object. So I am puzzled why this change was made 2 weeks ago.

I use python 3.5.2

ptnapoleon:
Fixed in  5143a76213f12cb114c561f96fff28881ef3ddd4 thanks!

PR #571, opened 2016-12-27
next can not be called on a TarFile object

ekooiker:
Addendum #556 (commit 5143a76)


ptnapoleon:
Huh, I'm not sure how that re-broke? But, thank you!

PR #573, opened 2017-01-11
Refactor callsite usage of check_socket_available

djbowers:


guyboltonking:
This request is based on the comment in https://github.com/pcmanus/ccm/pull/461/files#r55258123.

mambocab:
Guy makes great points -- the use case for `check_` returning a `bool` is limited, and this is more a case of bad naming than a bad API. Big +1 to Guy's "revert, change error message, `s/check/assert/`" plan.

ccmlib/common.py
```diff
@@ -492,12 +492,10 @@ def check_socket_available(itf, return_on_error=False):
...
         s.close()
         return True
-    except socket.error as msg:
+    except socket.error:
         s.close()
-        if return_on_error:
-            return False
         addr, port = itf
-        raise UnavailableSocketError("Inet address %s:%s is not available: %s" % (addr, port, msg))
+        raise UnavailableSocketError("Inet address %s:%s is not available; a cluster may already be running or you may need to add the loopback alias" % (addr, port))
```
guyboltonking: Nit: I'd keep the msg param, just for completeness (in case the reason for failure is something different):

```
"Inet address %s:%s is not available (%s); a cluster may already be running or you may need to add the loopback alias" % (addr, port, msg)
```

guyboltonking:
Apart from the above nit, great! +1

PR #574, opened 2017-01-12
Increase waiting time for protocols to 30 seconds

thibaultcha:
Quite often, it takes 15 or 20 seconds for the CQL binary protocol to start listening when starting a node. I thought about adding a `--protocol-wait-timeout` option but I think the added value is very limited. Simply increasing the timeout to 30 seconds will probably handle most cases.


thibaultcha:
Actually, a configurable value is probably more future-proof.

PR #577, opened 2017-01-30
Adding the ability to send SIGHUP to cluster and node(s) when stopping

mikefero:


ptnapoleon:
-1? This breaks the node stop API. If it would be nice if we still accepted `gently` (perhaps in **kwargs), and then simply set signal_event to KILL if gently is False.

mikefero:
I thought there might be some friction due to the API changes; however this change does allow us to easily send different signals when stopping the cluster/node(s). I am not 100% against your approach (and here it comes), but to use this new feature on the command line we would need to combine `--not-gently` and `--hang-up` in this example. Again not against it just would like to  get your (or anyone) thoughts on this.

If we go with the modified API changes we would definitely want to bump up the minor version number to indicate breaking API changes to the rest of the community.

ptnapoleon:
I'm not sure I agree? The CLI you have proposed right now is 100% fine, regardless of what we do with the python API. There's no need for a CLI user to care what kwarg `--not-gently` was setting before, as long as both work out to a SIGKILL. I'm thinking stop looks something like:
```
def stop(self, wait=True, signal_event=signal.SIGTERM, **kwargs):
    if 'gently' in kwargs and kwargs['gently'] is False:
        signal_event = signal.SIGKILL
        logger.warn("Overriding signal_event because gently=False was passed")
   ...
```

Thoughts?

mikefero:
Gotcha! I was under the assumption of reverting the parser options back (e.g. https://github.com/pcmanus/ccm/pull/577/files#diff-bfc8e41373dbb1f6daa37d16bbc253ecL619) and since in that scenaio `gently` would default to **`true`** the combined/nested paramters for the command line would be required if we wanted to perform more than just a `SIGKILL`.

This should work; let me whip something up and run a few tests.

mikefero:
Applied change requests and ensured all calls to `stop` where using keywords instead of being positional arguments.

Squashed and ready for review.

ptnapoleon:
+1, thanks!

PR #579, opened 2017-02-07
Correct signal events on Windows

mikefero:
Allowable Windows signals can be found here: https://msdn.microsoft.com/en-us/library/xdkz3x12.aspx#Anchor_2


ptnapoleon:
+1

PR #581, opened 2017-02-08
Use PID as part of tmp file name to avoid race conditions 

hkroger:
... of running operations in parallel. 

They would end up creating same conf anyways, right?


ptnapoleon:
At the very least, this doesn't seem like it can hurt, so +1

PR #582, opened 2017-02-13
Handle branch divergences in git cache more appropriately

ptnapoleon:
@jkni (and @djbowers ) to review please. I'll run some CI to make sure this doesn't break anything.


ptnapoleon:
http://cassci.datastax.com/view/Dev/view/ptnapoleon/job/ptnapoleon-test-trunk-dtest/3/console

djbowers:
Looks good to me. But I did find what looks like a type above the new code you wrote. Line 210 looks like it has an extra open bracket. 
`if str(status).find('[behind') > -1:`

ptnapoleon:
It is not a typo. At that point the value of `status` looks like
`'## cassandra-2.2...origin/cassandra-2.2 [behind 9]\n'`
and thus we expect the `[`. If we didn't have that, branches with the word `behind` in the name would always break

ptnapoleon:
Hard -1 on this, `up-to-date` is never printed in `status -sb`

ptnapoleon:
Okay, needs a new review

ptnapoleon:
http://cassci.datastax.com/view/Dev/view/ptnapoleon/job/ptnapoleon-test-trunk-dtest/4/

PR #584, opened 2017-02-15
Rotate & do not overwrite last.log

ptnapoleon:
I'm not particularly sure who might want to review this, let's try @guyboltonking?


ptnapoleon:
As mentioned, this is indisputably a mediocre way to do this, but unfortunately, the only other ones I could think of were worse

smccarthy788:
Good workaround for what we have. +1

PR #585, opened 2017-02-23
New Rotated CCM log generates too much trash

djbowers:


ptnapoleon:
Instead of doing this, just remove the calls to `rotate_log` on lines 151 and 350.

djbowers:
I have verified that this works and multiple log files are created. There may be more work we want to do on it, like grouping all of the repeated code into a function, but this is a good place to start.

ptnapoleon:
So, it looks like you could move
```
out = process.wait()
stdoutdata, stderrdata = process.communicate()
logger.info(stdoutdata)
logger.info(stderrdata)
```
into a function that takes process [and calls get_logger() itself] and returns the return code. That would remove a lot of the verbosity.

djbowers:
So I didn't put the call to get_logger() in the new function because the logger is used outside of that function as well (such as in line 370). Either way it doesn't save us an argument because we either have to pass logfile or logger to log_info().

ccmlib/repository.py
```diff
@@ -512,13 +533,15 @@ def lastlogfilename():
...
-    """
+def get_logger(log_file):
     logger = logging.getLogger('repository')
     logger.addHandler(handlers.RotatingFileHandler(log_file, maxBytes=1024*1024*5, backupCount=5))
-    logger.debug("Checking if we can rotate.")
+    return logger
+
+
+def log_info(process, logger):
+    out = process.wait()
```
ptnapoleon: process.wait() can deadlock. Instead you should do
```

    out, err = process.communicate()
    rc = process.returncode
```
and then return `rc`

ccmlib/repository.py
```diff
@@ -540,8 +540,8 @@ def get_logger(log_file):
 
 
 def log_info(process, logger):
-    out = process.wait()
+    rc = process.returncode
     stdoutdata, stderrdata = process.communicate()
```
ptnapoleon: You'll need to call communicate and _then_ check the returncode. Sorry for being wrong in my last comment

djbowers: Nope that makes sense I just wasn't thinking.

ptnapoleon:
+1, but let's run some CI to verify

ccmlib/repository.py
```diff
@@ -158,14 +158,14 @@ def clone_development(git_repo, version, verbose=False, alias=False):
             process = subprocess.Popen(
                 ['git', 'clone', '--mirror', git_repo, local_git_cache],
                 cwd=__get_dir(), stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-            out = log_info(process, logger)
+            out = log_info(process, logger)[0]
```
ptnapoleon: So, it's actually more idiomatic python to write:
`out, _, _ = log_info(process, logger)`,

but otherwise, this looks like the right fix

ptnapoleon:
+1

PR #587, opened 2017-03-15
Update dsetool to use node address and port

tolbertam:
dsetool previously always targeted localhost.  Since dsetool is a
node-specific command, it makes sense to use the ip address of the node.
In addition, also passes in the binary port to use in the event that it
is different than the default port (9042).

### DIFF
--- a/ccmlib/dse_node.py
+++ b/ccmlib/dse_node.py
@@ -175,8 +175,9 @@ class DseNode(Node):
     def dsetool(self, cmd):
         env = self.get_env()
         extension.append_to_client_env(self, env)
+        (node_ip, binary_port) = self.network_interfaces['binary']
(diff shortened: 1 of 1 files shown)


ccmlib/dse_node.py
```diff
@@ -175,8 +175,9 @@ def nodetool(self, cmd, username=None, password=None, capture_output=True, wait=
     def dsetool(self, cmd):
         env = self.get_env()
         extension.append_to_client_env(self, env)
+        (node_ip, binary_port) = self.network_interfaces['binary']
```
ptnapoleon: Just a nit, there shouldn't be parentheses here.

ptnapoleon:
+1 after nit

tolbertam:
good call, those definetely aren't needed, fixed

PR #588, opened 2017-03-21
Fixed python3 compatibility issue

bjmb:
Looks like status is a string looking like
`b'## cassandra-3.0...origin/cassandra'` 
which causes re.search to fail in python3 in an error that looks like 
`can't use a string pattern on a bytes-like object`


ptnapoleon:
Thanks!

PR #589, opened 2017-03-22
Allow sstabledump to print results from command

mike-tr-adamson:
This adds an additional parameter to the `run_sstabledump` method to allow the command to print it's results immediately rather than returning them in an array.

This changes lets `ccm node1 -k ks -c tb` work correctly on a >3.0 node while maintaining compatibility with the dtests that also use `run_sstabledump`


ptnapoleon:
+1

PR #591, opened 2017-03-28
grep WARN log lines for exceptions

smccarthy788:
@ptnapoleon, for your consideration.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -1986,12 +1986,22 @@ def _grep_log_for_errors(log):
     loglines = log.splitlines()
 
     for line_num, line in enumerate(loglines):
-        if log_line_category(line) == 'ERROR':
+        found_exception = False
+        line_category = log_line_category(line)
+        if line_category == 'ERROR':
             matches.append([line])
+            found_exception = True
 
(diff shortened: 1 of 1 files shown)


ccmlib/node.py
```diff
@@ -1986,12 +1986,22 @@ def log_line_category(line):
...
     for line_num, line in enumerate(loglines):
-        if log_line_category(line) == 'ERROR':
+        found_exception = False
+        line_category = log_line_category(line)
+        if line_category == 'ERROR':
             matches.append([line])
+            found_exception = True
 
+        elif line_category == 'WARN':
+            match = re.search(r'exception', line)
```
ptnapoleon: I don't think this will work. We're going to append a `MatchObject` to matches, but in the `ERROR` block, we append strings.

smccarthy788: line 1996 is just checking if line contains `'exception'`. The actual appending happens on lines 1992 and 1998: `matches.append([line])`

ptnapoleon: Woof, it totally does work like that. Could you make that `if match is not None`? I think they're functionally identical, but it matches the API for re.search more identically

PR #593, opened 2017-04-03
Closed open socket in check_socket_listening

bjmb:
Hi,

Sometimes the socket is left opened here. I'm seeing a bunch of these logs in python3 every time the function is called:

`ResourceWarning: unclosed <socket.socket fd=5, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('0.0.0.0', 43928)>`


ptnapoleon:
Awesome fix, +1

PR #596, opened 2017-05-11
Fix opscenter download

snazy:
This PR corrects Opscenter download location to `enterprise` instead of `community` and uses download credentials.


ptnapoleon:
+1, not sure what travis is mad about

PR #604, opened 2017-06-08
Removed broken tests

MichaelHamm:
Removed broken tests, tox, and added nosetests to the travis config file.


ptnapoleon:
It looks like there are still test files that are failing in the travis run. Go ahead and remove those too please

PR #606, opened 2017-06-22
handle failure in dse shell-out methods

mambocab:
This adds the handling added in #510 to DSE methods.

It's not clear to me how the C\* counterparts in `node.py` are tested. What should I do to test these changes? They work locally for me.

This is a breaking API change.

### DIFF
--- a/ccmlib/dse_node.py
+++ b/ccmlib/dse_node.py
@@ -13,7 +13,8 @@ import yaml
 from six import iteritems, print_
 
 from ccmlib import common, extension, repository
-from ccmlib.node import Node, NodeError, ToolError
+from ccmlib.node import (Node, NodeError, ToolError,
+                         handle_external_tool_process)
 
 
 class DseNode(Node):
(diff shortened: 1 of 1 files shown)


mambocab:
Oh, side note: the `handle_external_tool_process` requires an argument list in addition to the process being handled, because `Popen` objects don't persist this data [until the addition of `Popen.args` in Python 3.3](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.args). 

One way to slightly improve this function would be to use a custom `Popen` subclass as suggested in [this StackOverflow answer](https://stackoverflow.com/questions/33258684/python-how-to-get-command-run-with-subprocess-popen). I'd be happy to throw this together for this PR.

mambocab:
Just fixed this -- `Popen` objects weren't redirecting output; this fixes that.

ptnapoleon:
> One way to slightly improve this function would be to use a custom Popen subclass as suggested in this StackOverflow answer. I'd be happy to throw this together for this PR.

This seems unnecessary? Unless it's more than just ugly that we're passing in the list of args as well

mambocab:
> This seems unnecessary? Unless it's more than just ugly that we're passing in the list of args as well

Nope, just an unnecessarily ugly API. For now, there's nothing it prevents us from doing.

PR #607, opened 2017-06-23
Node now supports the configuration of a byteman startup script, to i…

sbtourist:
…nject behaviour at startup time.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -74,7 +74,7 @@ class Node(object):
     Provides interactions to a Cassandra node.
     """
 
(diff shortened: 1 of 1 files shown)


ccmlib/node.py
```diff
@@ -1610,6 +1611,8 @@ def __update_envfile(self):
         if self.byteman_port != '0':
             byteman_jar = glob.glob(os.path.join(self.get_install_dir(), 'build', 'lib', 'jars', 'byteman-[0-9]*.jar'))[0]
             agent_string = "-javaagent:{}=listener:true,boot:{},port:{}".format(byteman_jar, byteman_jar, str(self.byteman_port))
+            if (self.byteman_startup_script != None):
```
ptnapoleon: Please change this to `if self.byteman_startup_script is not None`

sbtourist: Done.

PR #611, opened 2017-07-26
Overwrite in.sh on upgrade. Closes #463

### LINKED ISSUE #463: Upgrade / Downgrade of node leaves previous jamm jar version in configs
While adding sstableupgrade_test in dtest, I found that when the jamm jar version changes between 2.0 and 2.1 Cassandra branches, CCM does not handle the configuration updates fully during upgrade/downgrade of a node, leaving the old jamm jar version in the configs. I'll work on repro steps and add them here, but I believe that this comes from the env setup in ccmlib/common.py.

ptnapoleon:
Testing locally before merging. @iamaleksey to review, please

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -246,8 +246,7 @@ def make_cassandra_env(install_dir, node_path, update_conf=True):
         sh_file = os.path.join(BIN_DIR, CASSANDRA_SH)
     orig = os.path.join(install_dir, sh_file)
     dst = os.path.join(node_path, sh_file)
-    if not os.path.exists(dst):
-        shutil.copy(orig, dst)
+    shutil.copy(orig, dst)
 
     if update_conf and not (is_win() and get_version_from_build(node_path=node_path) >= '2.1'):
         replacements = [


ptnapoleon:
I've run CI on this, and nothing fails that wasn't failing already. Going to go ahead and merge this w/o waiting for US reviewers to wake up.

bjmb:
Looks like this commit is making our tests in appveyor fail: https://ci.appveyor.com/project/DataStax/python-driver/build/1528/job/j82bjrm0cwvq4a8o#L1479 . 
We can start a cluster and remove it but when we try to start it again there's an error starting the nodes.
Seems fine in linux.

ptnapoleon:
Damn. :(. We can.... gate the if check on OS? That will leave in the upgrade bug on windows, but fix the issue you're seeing?

bjmb:
Yeah I think that would be a solution for us, also allowing this might be OK since probably not that many people would run into the bug on Windows.

bjmb:
Thanks, they are back to normal.

PR #613, opened 2017-08-02
Print output from dsetool. Don't capture dse interactive shells.

ptnapoleon:
@mambocab and @bcantoni to review, please

### DIFF
--- a/ccmlib/cmds/node_cmds.py
+++ b/ccmlib/cmds/node_cmds.py
@@ -370,7 +370,9 @@ class _DseToolCmd(Cmd):
         Cmd.validate(self, parser, options, args, node_name=True, load_cluster=True)
 
     def run(self):
-        self.node.dsetool(self.dsetool_cmd)
+        stdout, stderr, rc = self.node.dsetool(self.dsetool_cmd + " " + " ".join((self.args[1:])))
+        print_(stderr)
+        print_(stdout)
 
 
 class NodeDsetoolCmd(_DseToolCmd):
@@ -378,7 +380,9 @@ class NodeDsetoolCmd(_DseToolCmd):
     descr_text = "Run dsetool (connecting to node name)"
 
     def run(self):
-        self.node.dsetool(" ".join(self.args[1:]))
+        stdout, stderr, rc = self.node.dsetool(" ".join(self.args[1:]))
+        print_(stderr)
+        print_(stdout)
 
 
 class NodeCliCmd(Cmd):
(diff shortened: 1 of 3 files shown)


ccmlib/dse_node.py
```diff
@@ -192,8 +192,8 @@ def dse(self, dse_options=None):
         dse = common.join_bin(self.get_install_dir(), 'bin', 'dse')
         args = [dse]
         args += dse_options
-        p = subprocess.Popen(args, env=env, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-        return handle_external_tool_process(p, args)
+        p = subprocess.Popen(args, env=env)  # Spawn an interactive shell, cannot pipe stdout/std err
```
mambocab: I don't think this comment is quite right -- it doesn't spawn a new shell, it spawns a new process, and that process inherits `std{in,out} from the parent `ccm` process. When called from a shell, this is just inherited from the shell's `std{in,out}`, so users can interact with the, e.g., `dse` process.

The comment is basically explaining the absence of `stdout` and `stdin` arguments, yeah? It took me a minute to get that. I'd rephrase as "don't redirect stdout/in so users can interact with new process" or something similar.

ccmlib/cmds/node_cmds.py
```diff
@@ -370,15 +370,19 @@ def validate(self, parser, options, args):
...
+        print_(stdout)
 
 
 class NodeDsetoolCmd(_DseToolCmd):
     usage = "usage: ccm node_name dsetool [options]"
     descr_text = "Run dsetool (connecting to node name)"
 
     def run(self):
-        self.node.dsetool(" ".join(self.args[1:]))
+        stdout, stderr, rc = self.node.dsetool(" ".join(self.args[1:]))
```
mambocab: The `nodetool` equivalent for this looks like ```stdout, stderr, rc = self.node.nodetool(self.nodetool_cmd + " " + " ".join((self.args[1:])))
```
So, it prepends the `*_cmd` attribute, where the `dsetool` version doesn't have that. I see that it's been that way, but it threw me off.

ccmlib/dse_node.py
mambocab: Ah, just saw the bug report for this. So, I know that `spark` is an interactive command -- are all the others?

The issue I see here is that `handle_external_tool_process` will raise an exception if the `returncode` is nonzero, which I think is the behavior we want for most of these commands. Maybe you want to make a change to the handler method that changes in/out behavior when input/output isn't redirected?

Issue #565, opened 2016-11-20
ccm stop - UnboundLocalError: local variable 'pidfile' referenced before assignment


dins2k2:
Hi,

ccm stop gives this error **UnboundLocalError: local variable 'pidfile' referenced before assignment**

```
dinesh@ccm01:~$ ccm create --dse -v 5.0.3 --dse-username=DSEUsername --dse-password=DSEPassword -o 5.2.4 -n 2:1 casops1
Current cluster is now: casops1
dinesh@ccm01:~$ ccm start
dinesh@ccm01:~$ ccm status
Cluster: 'casops1'
------------------
node1: UP
node3: UP
node2: UP
dinesh@ccm01:~$ ccm stop
Traceback (most recent call last):
  File "/usr/local/bin/ccm", line 5, in <module>
    pkg_resources.run_script('ccm==2.4.7', 'ccm')
  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 528, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 1401, in run_script
    exec(script_code, namespace, namespace)
  File "/usr/local/lib/python2.7/dist-packages/ccm-2.4.7-py2.7.egg/EGG-INFO/scripts/ccm", line 90, in <module>

  File "build/bdist.linux-x86_64/egg/ccmlib/cmds/cluster_cmds.py", line 630, in run
  File "build/bdist.linux-x86_64/egg/ccmlib/dse_cluster.py", line 76, in stop
  File "build/bdist.linux-x86_64/egg/ccmlib/cluster.py", line 435, in stop
  File "build/bdist.linux-x86_64/egg/ccmlib/dse_node.py", line 146, in stop
  File "build/bdist.linux-x86_64/egg/ccmlib/dse_node.py", line 153, in _stop_agent
UnboundLocalError: local variable 'pidfile' referenced before assignment
```

Ubuntu 14.04, Python 2.7

Thanks,
Dinesh

PR #616, opened 2017-08-14
Fix indentation in _stop_agent

sproberts92:
Commit `f9c70c0` introduced an indentation error whereby the variable
pidfile was referenced without being defined. This solves issue #565, where
`ccm stop` fails upon calling `_stop_agent()`.


sproberts92:
Travis has failed on downloading Python 2.7 binaries, not related to my commit. Jobs for 3.4.5 and 3.5.2 have passed ok.

PR #618, opened 2017-08-24
Allow options when calling compact, flush, cleanup API

### LINKED ISSUE #546: let compact take options
It's sort of surprising that `node.Node.compact` doesn't take any options and can only ever call `nodetool compact`.

https://github.com/pcmanus/ccm/blob/master/ccmlib/node.py#L1281-L1282

It might be nice to flesh this out a bit, if only to take a list of options.

hatokani2:
followed the implementation of "repair function".
close #546


ptnapoleon:
Great PR, Thanks @hatokani2!

PR #620, opened 2017-08-30
Prohibit non-word characters as clustername

cormoran:
Only allow `[a-zA-Z0-9_-]+` in the cluster name.
Please see  https://github.com/pcmanus/ccm/issues/421.


ptnapoleon:
Thanks for the fix!

PR #621, opened 2017-08-30
add Node.get_sstables_via_sstableutil

snazy:


ccmlib/node.py
```diff
@@ -1249,6 +1249,28 @@ def get_sstables_per_data_directory(self, keyspace, column_family):
...
+            args.extend(['--cleanup'])
+
+        args.extend([keyspace, table])
+
+        p = subprocess.Popen(args, env=env, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+
+        (stdout, stderr) = p.communicate()
+
+        if p.returncode != 0:
+            raise Exception("Error invoking sstableutil; returned {code}".format(code=p.returncode))
```
ptnapoleon: We should log the stdout and stderr when this happens

ccmlib/node.py
```diff
@@ -1249,6 +1249,28 @@ def get_sstables_per_data_directory(self, keyspace, column_family):
...
+        if oplogs:
+            args.extend(['--oplog'])
+        if cleanup:
+            args.extend(['--cleanup'])
+
+        args.extend([keyspace, table])
+
+        p = subprocess.Popen(args, env=env, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+
+        (stdout, stderr) = p.communicate()
```
ptnapoleon: No need for the parentheses here

ccmlib/node.py
```diff
@@ -1249,6 +1249,28 @@ def get_sstables_per_data_directory(self, keyspace, column_family):
     def get_sstables(self, keyspace, column_family):
         return [f for sublist in self.get_sstables_per_data_directory(keyspace, column_family) for f in sublist]
 
+    def get_sstables_via_sstableutil(self, keyspace, table, type='all', oplogs=False, cleanup=False, match = '-Data.db'):
```
ptnapoleon: I don't think there should be spaces around the `=` after `match`

ptnapoleon: `type` is a keyword in python, best not to override it here.

PR #622, opened 2017-08-31
Support byteman startup scripts when calling start()

ostefano:
I am working on a new test in cassandra-dtest where I need to submit a byteman rule to a single node,  after restarting it. Might not be the best way, so I am open to suggestions.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -563,7 +563,8 @@ class Node(object):
               use_jna=False,
               quiet_start=False,
               allow_root=False,
-              set_migration_task=True):
+              set_migration_task=True,
+              byteman_startup_script=None):
         """
         Start the node. Options includes:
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Hi @ostefano. Forgive me, but I don't really know how the byteman integration in C* works. Why can't you simply set `node.byteman_startup_script` before calling `start()`?

ostefano:
Hi @ptnapoleon , that would not work because `byteman_startup_script` needs to be part of the configuration file, and setting the class member does not update it (it is only updated when `__update_envfile` is called by the node constructor).

ptnapoleon:
Okay. Then, for code re-uses sake, maybe let's just add a method `node.update_startup_byteman_script` that changes the class member and runs __update_envfile (or import_config_files)?

ostefano:
Sure, sounds good to me. Pushed commit e70cd78

PR #624, opened 2017-09-05
Derive ip address based on thrift or binary interface

tolbertam:
In dse_node there are some cases where the thrift network_interface is
used to derive the ip address.  It is not guaranteed that thrift will be
enabled/present.  In this case, create an ip_addr member that is set
based on thrift or binary interface ip address.

### DIFF
--- a/ccmlib/dse_node.py
+++ b/ccmlib/dse_node.py
@@ -49,8 +49,7 @@ class DseNode(Node):
         return [common.join_bin(os.path.join(self.get_install_dir(), 'resources', 'cassandra'), 'bin', 'dse'), toolname]
 
     def get_env(self):
-        (node_ip, _) = self.network_interfaces['binary']
-        return common.make_dse_env(self.get_install_dir(), self.get_path(), node_ip)
+        return common.make_dse_env(self.get_install_dir(), self.get_path(), self.ip_addr)
 
     def get_cassandra_version(self):
         return common.get_dse_cassandra_version(self.get_install_dir())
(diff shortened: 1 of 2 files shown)


ccmlib/dse_node.py
```diff
@@ -377,7 +375,7 @@ def __generate_server_xml(self):
         with open(server_xml, 'w+') as f:
             f.write('<Server port="8005" shutdown="SHUTDOWN">\n')
             f.write('  <Service name="Solr">\n')
-            f.write('    <Connector port="8983" address="%s" protocol="HTTP/1.1" connectionTimeout="20000" maxThreads = "200" URIEncoding="UTF-8"/>\n' % self.network_interfaces['thrift'][0])
+            f.write('    <Connector port="8983" address="%s" protocol="HTTP/1.1" connectionTimeout="20000" maxThreads = "200" URIEncoding="UTF-8"/>\n' % self.ip_addr)
```
tolbertam: This is the main issue for me, if thrift is not enabled and I attempt to use the solr workload, I get an error:

```
  File "/.../python2.7/site-packages/ccmlib/dse_node.py", line 380, in __generate_server_xml
    f.write('    <Connector port="8983" address="%s" protocol="HTTP/1.1" connectionTimeout="20000" maxThreads = "200" URIEncoding="UTF-8"/>\n' % self.network_interfaces['thrift'][0])
TypeError: 'NoneType' object has no attribute '__getitem__'
```

PR #625, opened 2017-09-06
Timeout in node start even when system.log is not written

ptnapoleon:
We were hanging forever if the jvm fails to start


mshuler:
+1

PR #629, opened 2017-09-19
removed nonexistent yaml key

xytxytxyt:
before:
```
ccm create test --install-dir=/Users/xianteng/bdp/
ccm populate -n 1
ccm setworkload graph
  File "/Users/xianteng/ctool-env/bin/ccm", line 90, in <module>
    cmd.run()
  File "/Users/xianteng/ctool-env/lib/python2.7/site-packages/ccmlib/cmds/cluster_cmds.py", line 855, in run
    node.set_workloads(workloads=self.workloads)
  File "/Users/xianteng/ctool-env/lib/python2.7/site-packages/ccmlib/dse_node.py", line 71, in set_workloads
    graph_options['gremlin_server']['host'] = self.ip_addr
TypeError: 'NoneType' object does not support item assignment
```
after, things work fine

it doesn't look like there was ever a `graph.gremlin_server.host` in dse.yaml

### DIFF
--- a/ccmlib/dse_node.py
+++ b/ccmlib/dse_node.py
@@ -64,12 +64,6 @@ class DseNode(Node):
         if 'solr' in self.workloads:
             self.__generate_server_xml()
         if 'graph' in self.workloads:
-            conf_file = os.path.join(self.get_path(), 'resources', 'dse', 'conf', 'dse.yaml')
-            with open(conf_file, 'r') as f:
-                data = yaml.load(f)
-            graph_options = data['graph']
-            graph_options['gremlin_server']['host'] = self.ip_addr
-            self.set_dse_configuration_options({'graph': graph_options})
             self.__update_gremlin_config_yaml()
         if 'dsefs' in self.workloads:
             dsefs_options = {'dsefs_options': {'enabled': True,


ccmlib/dse_node.py
```diff
@@ -64,12 +64,6 @@ def set_workloads(self, workloads):
         if 'solr' in self.workloads:
             self.__generate_server_xml()
         if 'graph' in self.workloads:
-            conf_file = os.path.join(self.get_path(), 'resources', 'dse', 'conf', 'dse.yaml')
-            with open(conf_file, 'r') as f:
-                data = yaml.load(f)
-            graph_options = data['graph']
-            graph_options['gremlin_server']['host'] = self.ip_addr
```
ptnapoleon: Why do we need to remove more than just this line? It would seem like the rest of this code is still functional

xytxytxyt: it looks like graph.gremlin_server.host is the only value changed, and with it gone, there's no need to read the yaml anymore?

newkek: +1 to @xytxytxyt

tolbertam:
Encountered this same issue and this seems to fix it 👍

PR #630, opened 2017-09-26
Add the ability to remotely execute CCM commands

mikefero:
* Handles `--dse-credentials`, `--node-ssl`, and `-ssl`
  * Files will be copied from local to remote before execution
* Updated documentation to reflect new remote options

This update adds the ability to specify a remote host to execute CCM
commands on. The remote machine must be pre-configured to accept SSH
connections, sequential IP addresses for simple node creation, and a
installed instance of CCM; see misc/Vagrantfile for simplified guest
installation.

__Note__: This update is optional and requires that `paramiko` is
          installed in order to function.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ python:
   - "3.4.5"
   - "3.5.2"
 install:
-  - pip install pylint six pyyaml
+  - pip install pylint six pyyaml paramiko
 script:
   - pylint --py3k --disable=W1633,W1648 ccmlib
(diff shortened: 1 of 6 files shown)


mikefero:
Tested using Windows 10 host against Ubuntu 16.04 guest
  * Windows 10 host used both Python v2.7.14 and v3.6.2
  * Ubuntu 16.04 used Python v2.7.12

mikefero:
Apparently I was squashing it as you were reviewing this morning (sorry about that). Thanks for the review; merge away.

PR #637, opened 2017-10-26
Configure and create SPARK_EXECUTOR_DIRS

tolbertam:


ccmlib/common.py
```diff
@@ -316,7 +316,11 @@ def make_dse_env(install_dir, node_path, node_ip):
     env['SHARK_CONF_DIR'] = os.path.join(node_path, 'resources', 'shark', 'conf')
     env['GREMLIN_CONSOLE_CONF_DIR'] = os.path.join(node_path, 'resources', 'graph', 'gremlin-console', 'conf')
     env['SPARK_WORKER_DIR'] = os.path.join(node_path, 'spark', 'worker')
-    env['SPARK_LOCAL_DIRS'] = os.path.join(node_path, 'spark', 'rdd')
+    if version >= '6.0':
+        env['SPARK_LOCAL_DIRS'] = os.path.join(node_path, 'spark', '.local')
+        env['SPARK_EXECUTOR_DIRS'] = os.path.join(node_path, 'spark', 'rdd')
```
JeremiahDJordan: I would just set this? no 6.0 check? like it doesn't hurt to have extra env vars, and putting 6.0 in here gives away our next DSE version number ;)

tolbertam: Good point, it doesn't hurt for it to be there and it'd simplify the code in dse_node.py to not have to check to see if the variable is set before creating the directory.   We already have some 6.0 specific checks in here so I figured the cat was out of the bag 😄 .

PR #641, opened 2017-11-17
Update node version on install_dir change

spodkowinski:
Is there any reason not to update the version value when changing the install directory for a node ([just as we do for cluster](https://github.com/pcmanus/ccm/blob/master/ccmlib/cluster.py#L96))? Otherwise this seems to cause issues for upgrade tests by sharing hints directories as described in 10f3ea95dee352a106bcb20182ec31354c490fc0.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -243,6 +243,8 @@ class Node(object):
         else:
             self.__install_dir = self.node_setup(version, verbose=verbose)
 
+        self._cassandra_version = common.get_version_from_build(self.__install_dir, cassandra=True)
+
         if self.get_base_cassandra_version() >= 4.0:
             self.network_interfaces['thrift'] = None
 


ptnapoleon:
No reason that I know of. Have you tested this with the upgrade tests, to confirm it works?

spodkowinski:
Still trying to get some of the tests working, it's definitely a combination of issues.

ptnapoleon:
Ahh, so at the very least, this doesn't break them more? If so, then I'm fine merging it.

spodkowinski:
I don't think I'm able to do a dtest run on builds.apache.org with a custom CCM build, so I haven't been able to verify how this may effect other related tests. But no need to rush to commit this from my side, I guess there'll be more patches coming anyways.

ptnapoleon:
Alright then, I'll run them locally over the weekend and merge on Monday. Thanks!

PR #645, opened 2017-12-03
Java 9 additions

snazy:
Fixes parsing of the version number (if the version number is just `9` or `10` for the `.0.0` releases).

Updates to GC-logging config for Java 9 and newer (applies to Java9 branch)

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -700,9 +700,19 @@ def invalidate_cache():
 
 
 def get_jdk_version():
+    """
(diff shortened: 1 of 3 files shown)


ptnapoleon:
Looks good, but I'm going to do a very mean thing. Can you please add a test for the get_jdk_version method? That covers jdk8 and jdk9 syntaxes for the version. Thank you

snazy:
Okay, added a unit test

PR #664, opened 2018-02-16
Decode output of external tool subprocess

spodkowinski:
Dtests interacting with external cassandra tools currently fail in Python3 environments, as the output is not decoded before passed to `JSON.loads()`.

Example:
https://builds.apache.org/view/A-D/view/Cassandra/job/Cassandra-devbranch-dtest/494/testReport/offline_tools_test/TestOfflineTools/test_sstabledump/

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -2099,6 +2099,8 @@ def _grep_log_for_errors(log):
 
 def handle_external_tool_process(process, cmd_args):
     out, err = process.communicate()
+    if out and isinstance(out, bytes):
+        out = out.decode()
     rc = process.returncode
 
     if rc != 0:


ptnapoleon:
It appears `out` isn't always of type `bytes` here, as indicated by Travis.

spodkowinski:
What would be the canonical way to deal with this in python?
1. `if isinstance(out, str): out.decode()`? Would not have any effect on python2, but which is probably ok as the python2 code should already work anyways
2. `if out.decode: out.decode()`

Also, I think we should not use the 'utf-8' literal and use the platform default encoding instead?

ptnapoleon:
`if out and isinstance(out, bytes): out.decode()`, I think is what we want? I don't know what's better between utf-8 and platform default.

spodkowinski:
That seems to fix it, thanks!

PR #666, opened 2018-02-19
Decode also err output of external tool subprocess

spodkowinski:
Byte values for err channel should be decoded as well, see 0a5824508

cc @aweisberg


aweisberg:
+1 in for a penny in for a pound.
So who is going to update the dtests instances where we decode?

ptnapoleon:
Is this safe to merge or do you want me to wait for some corresponding dtest changes?

aweisberg:
it's not safe, but a lot of these instances are already broken on decoding stdout anyways.
So... umm... sure merge it?

PR #667, opened 2018-02-21
Grep log entry category as word

snazy:


ccmlib/node.py
```diff
@@ -2029,7 +2029,7 @@ def _get_load_from_info_output(info):
 
 def _grep_log_for_errors(log):
     except_re = re.compile(r'[Ee]xception|AssertionError')
-    log_cat_re = re.compile(r'(INFO|DEBUG|WARN|ERROR)')
+    log_cat_re = re.compile(r'\W(INFO|DEBUG|WARN|ERROR)\W')
```
ptnapoleon: I don't think \W matches start of line

PR #675, opened 2018-05-08
Add ability to configure archive repository locations via ~/.ccm/config

tolbertam:
This adds the capability to override the repository locations for downloading cassandra, DSE, and OpsCenter.   This is useful for situations such as when CCM is used as part of a continuous integration workflow where a local mirror is set up so downloads can be done quickly over the local network.

Repositories can be configured via the `[repositories]` section of the `~/.ccm/config` file which was already defined for git aliases.  For example:

```
[repositories]
cassandra = http://mirror.awesomeco.intranet/apache/cassandra
dse = http://mirror.awesomeco.intranet/tar/dse/dse-%s-bin.tar.gz
opscenter = http://mirror.awesomeco.intranet/tar/opscenter/opscenter-%s.tar.gz
```

### DIFF
--- a/ccmlib/repository.py
+++ b/ccmlib/repository.py
@@ -269,7 +269,11 @@ def clone_development(git_repo, version, verbose=False, alias=False):
 
 
 def download_dse_version(version, username, password, verbose=False):
-    url = DSE_ARCHIVE % version
+    url = DSE_ARCHIVE
+    if CCM_CONFIG.has_option('repositories', 'dse'):
+        url = CCM_CONFIG.get('repositories', 'dse')
+
+    url = url % version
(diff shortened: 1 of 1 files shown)


ccmlib/repository.py
```diff
@@ -327,10 +335,14 @@ def download_version(version, url=None, verbose=False, binary=False):
     """
     assert_jdk_valid_for_cassandra_version(version)
 
+    u = ARCHIVE
```
mikefero: What do you think about renaming `u` => `url`?

tolbertam: There is already a `url` kwarg to the function,  I can come up with a better name though.

PR #679, opened 2018-07-06
Use RawConfigParser instead of ConfigParser

tolbertam:
ConfigParser supports string interpolation which I never intended
to be supported and causes issues with Python 3.x when specifying
a dse mirror repository, which expects a format like:

[repositories]
dse = http://myrepo/enterprise/dse-%s-bin.tar.gz

Unfortunately, ConfigParser interpolation is not compatible between
2.x and 3.x as 3.x requires escaping '%', where 2.x does not unescape
'%', so using RawConfigParser seems like the best viable solution.

### DIFF
--- a/ccmlib/dse_cluster.py
+++ b/ccmlib/dse_cluster.py
@@ -53,7 +53,7 @@ class DseCluster(Cluster):
                 dse_credentials_file = creds_file
 
         if dse_credentials_file is not None:
-            parser = ConfigParser.ConfigParser()
+            parser = ConfigParser.RawConfigParser()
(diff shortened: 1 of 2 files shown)


tolbertam:
Note:  This impacts both `~/.ccm/config` (currently only used for mirror locations) and ~/.ccm/.dse.ini`  (dse credentials).  I figured I'd update both to be consistent and made sense to do for .dse.ini as well since it's not uncommon for people to use `%` in their passwords.

tolbertam:
Thanks for the quick merge! :)

PR #680, opened 2018-07-23
CASSANDRA-9608 upgrade all jvm.options files

snazy:
CASSANDRA-9608 splits the `jvm.options` file into multiple files. The patch makes ccm check all those files.


snazy:
I can confirm, that the change works against current trunk (i.e. the single `jvm.options` file) and 9608 (multiple files).

PR #682, opened 2018-08-23
Add warning when CASSANDRA_HOME is set in environment

nattochaduke-yahoo:
Solution for Issue[340](https://github.com/riptano/ccm/issues/340).
If there is CASSANDRA_HOME environment variable, it will be warned.

### DIFF
--- a/ccm
+++ b/ccm
@@ -8,6 +8,9 @@ from ccmlib import common
 from ccmlib.cmds import command, cluster_cmds, node_cmds
 from ccmlib.remote import PARAMIKO_IS_AVAILABLE, get_remote_usage, get_remote_options, execute_ccm_remotely
 
+import os
+import warnings
+
 
 def get_command(kind, cmd):
     cmd_name = kind.lower().capitalize() + cmd.lower().capitalize() + "Cmd"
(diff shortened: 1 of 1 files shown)


ccm
```diff
@@ -92,6 +95,9 @@ else:
     cmd = sys.argv[2]
     cmd_args = [node] + sys.argv[3:]
 
+if not os.getenv('CASSANDRA_HOME') is None:
```
ptnapoleon: Could you rephrase this as `if os.getenv('CASSANDRA_HOME') is not None`, please? PEP-8 specifies that form is preferable to prefixing with `if not`

ccm
```diff
@@ -8,6 +8,9 @@ from ccmlib import common
 from ccmlib.cmds import command, cluster_cmds, node_cmds
 from ccmlib.remote import PARAMIKO_IS_AVAILABLE, get_remote_usage, get_remote_options, execute_ccm_remotely
 
+import os
+import warnings
```
ptnapoleon: This is the wrong place for these imports. You could install and run `isort ccm` against this file to have it automatically cleaned, but stdlib imports belong first. So:
```
import os
import sys
import warnings

import pkg_resources
...
```

nattochaduke-yahoo:
I am sorry I mistook handling branches and being confused.
I will refine my codes and push tommorw.

ptnapoleon:
Thank you!

Issue #372, opened 2015-09-02
`ccm multitail`


mambocab:
The other day I saw @nutbunnies using `multitail` to keep an eye on the log for every node in the cluster. It seemed real cool and really useful for debugging. This'd be a good ccm feature, I think. I think I'd also want a `CCM_MULTITAIL_CMD` option so users can use their preferred log-watching tool.

PR #683, opened 2018-08-28
Solution for #372 , Add "showlogs" command and CCM_MULTITAIL_CMD environment

nattochaduke-yahoo:
#372 Requested adding ccm option that shows nodes' information in a parallel manner.
We implemented that feature "showlogs". By default it shows every node's log by "multitail" command.
By exporting command and options to CCM_MULTITAIL_CMD, you can execute any command and options to view the logs. If you want to specify nodes to be shown in a cluster, execute like
$ ccm showlogs [node1, node2, ...]

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -705,7 +705,8 @@ class Cluster(object):
 
     def show_logs(self, selected_nodes_names=()):
         """
-        Shows logs of nodes in this cluster with multitail.
+        Shows logs of nodes in this cluster, by default, with multitail.
+        If you need to alter the command or options, change CCM_MULTITAIL_CMD .
         Params:
         @selected_nodes_names : a list-like object that contains names of nodes to be shown. If empty, this will show all nodes in the cluster.
 
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -702,3 +702,22 @@ class Cluster(object):
 
     def wait_for_any_log(self, pattern, timeout, filename='system.log', marks=None):
         return common.wait_for_any_log(self.nodelist(), pattern, timeout, filename=filename, marks=marks)
+
+    def show_logs(self, selected_nodes_names=()):
+        nodes = sorted(list(self.nodes.values()), key=lambda node: node.name)
+        nodes_names = [node.name for node in nodes]
+
+        if len(nodes) == 0:
+            print("No node in this cluster yet.")
+            return
+        else:
+            names_logs_dict = {node.name: node.logfilename() for node in nodes}
+            if len(selected_nodes_names) == 0:
+                return names_logs_dict.values()
+            else:
+                if set(selected_nodes_names).issubset(nodes_names):
+                    return [names_logs_dict[name] for name in selected_nodes_names]
+                else:
+                    raise ValueError("nodes in this cluster are {}. But nodes in argments are {}".format(
+                        nodes_names, selected_nodes_names
+                    ))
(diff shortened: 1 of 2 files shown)


ccmlib/cmds/cluster_cmds.py
```diff
@@ -345,6 +346,22 @@ def run(self):
         self.cluster.show(self.options.verbose)
 
 
+class ClusterShowlogsCmd(Cmd):
+    descr_text = "Show logs of nodes in this claster. If no nodes are specified, logs of all nodes will be shown.\
```
ptnapoleon: nit: s/claster/cluster/

ccmlib/cmds/cluster_cmds.py
```diff
@@ -345,6 +346,22 @@ def run(self):
...
+                 By default multitail is used. If you need to alter the command or options, change CCM_MULTITAIL_CMD."
+
+    usage = "usage: ccm showlogs [node1 node2 ...]"
+
+    def validate(self, parser, options, args):
+        Cmd.validate(self, parser, options, args, load_cluster=True)
+
+    def run(self):
+        shower = os.getenv("CCM_MULTITAIL_CMD", 'multitail').split()[0]
```
ptnapoleon: Can you remove all of the `CCM_MULTITAIL_CMD` env based configuration? I don't think we need to support multiple options.

ccmlib/cluster.py
```diff
@@ -702,3 +702,31 @@ def timed_grep_nodes_for_patterns(self, versions_to_patterns, timeout_seconds, f
 
     def wait_for_any_log(self, pattern, timeout, filename='system.log', marks=None):
         return common.wait_for_any_log(self.nodelist(), pattern, timeout, filename=filename, marks=marks)
+
+    def show_logs(self, selected_nodes_names=()):
```
ptnapoleon: This is a surprising python gotcha, but one should never use `()` or []` as the default value for a keyword argument. Everytime this method is called, you don't get a new empty tuple, but end up re-using the same tuple from previous calls. It is recommended to instead:
```
def show_logs(self, selected_node_names=None):

    if selected_node_names is None:
        selected_node_names = ()
....
```

ptnapoleon:
Thanks for the PR! I left some feedback. Could you make those changes, and also fix the issues causing the Travis CI check to fail? Thank you.

ccmlib/cmds/cluster_cmds.py
nattochaduke-yahoo: Thanks for your comment, however original issue #372 requests this feature and I believe that this feature is useful when the users do not have multitail in their systems or they want to flexibly change the arrangement of mini-panes (vertical split, horizontal split, etc).

ccmlib/cluster.py
nattochaduke-yahoo: OK, I fix this. Thank you.

ccmlib/cmds/cluster_cmds.py
nattochaduke-yahoo: OK, I fix this. Thank you.

ccmlib/cmds/cluster_cmds.py
ptnapoleon: Hmm, you're right, Jim did ask for that. Nevermind then.

ptnapoleon:
Thanks! P.S. If you include "Closes #XYZ" such as `Closes #372` in your PR title or a commit message, github automatically closes the matching issue.

PR #688, opened 2018-12-19
More fixes for CASSANDRA-14421 to make changing versions easier

aweisberg:
When setting a new version use LooseVersion for the version field in Cluster to be consistent. Some test code fails because it expects a LooseVersion all the time.

When going from > 4.0 version to a < 4.0 version remove the 4.0 options automatically.

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -11,6 +11,7 @@ import sys
 import threading
 import time
 from collections import OrderedDict, defaultdict, namedtuple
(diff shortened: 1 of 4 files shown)


ptnapoleon:
Could you fix the linting issues reported by Travis? Otherwise, lgtm

aweisberg:
I am not sure what those issues were, but it seems to be accepting the changes now. I didn't squash the build output change just because I think it would be a better history to keep them separate.

PR #703, opened 2019-11-27
allowing dash in Cassandra version number, i.e. 4.0.alpha2

### LINKED ISSUE #702: Can't create cluster using 4.0-alpha binary releases
According to the documentation you should be able to create a cluster using binary releases at http://archive.apache.org/dist/cassandra/

However, I get a failure when I try to create a cluster using "4.0-alpha2"

```
$ ccm create -v 4.0-alpha2 -n 3 my_cluster
17:50:53,154 ccm WARNING Downloading 4.0-alpha2 failed, trying to build from git instead.
The error was: Invalid version 4.0-alpha2 (underlying error is: HTTP Error 404: Not Found)
```

the 404 is a clue that it is not building the right path to the binary.

Looking at
https://github.com/riptano/ccm/blob/master/ccmlib/repository.py#L343, seems like the "-alpha" is getting stripped off in building the path to the binary:

```
        archive_url = "%s/%s/apache-cassandra-%s-bin.tar.gz" % (archive_url, version.split('-')[0], version) if url is None else url
```

So that `split` would mean it's trying to load:
http://archive.apache.org/dist/cassandra/4.0/apache-cassandra-4.0-alpha2-bin.tar.gz

instead of:
http://archive.apache.org/dist/cassandra/4.0-alpha2/apache-cassandra-4.0-alpha2-bin.tar.gz

I don't understand the purpose of having that split there.

jeffreyscarpenter:
fixes #702

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -281,7 +281,7 @@ class Node(object):
 
     def get_base_cassandra_version(self):
         version = self.get_cassandra_version()
-        return float('.'.join(version.vstring.split('.')[:2]))
+        return float('.'.join(re.split('\.|-',version.vstring)[:2]))
 
     def set_configuration_options(self, values=None):
         """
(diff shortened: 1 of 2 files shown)


ptnapoleon:
This looks good, but I'm running the dtests against this now, just to be sure. Looking at the list of versions on http://archive.apache.org/dist/cassandra, I dont see anyway this could be problematic

PR #712, opened 2020-04-20
ccmlib should not always terminate cql statements with a ';'

dineshjoshi:
I am running into an issue that ccmlib assumes [all input will be terminated by a `;`](https://github.com/riptano/ccm/blob/275699f79d102b5039b79cc17fa6305dccf18412/ccmlib/node.py#L910). This isn't true in case of lets say comments or multi-line comments. ccmlib should optionally take a terminator to avoid this issue.

Here's an example of a failure:

```python
self.node1.run_cqlsh(
   """/* multiline
    *
    * comment */
    """)
```

Attached patch fixes the issue. @snazy @nastra could you please review this PR?

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -902,7 +902,7 @@ class Node(object):
         p = self.verify_process(options=options)
         return handle_external_tool_process(p, ['sstableverify'] + options)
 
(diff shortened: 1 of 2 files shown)


dineshjoshi:
thank you.

PR #713, opened 2020-06-02
CASSANDRA-15835: Use the right Java version + JAVA_HOME when building C* from source / cassandra-test

snazy:
Also:
* Add a new "version source type" `clone:/path/to/local/cassandra/clone` to re-enable upgrade-dtests for dev-CI and local upgrade-dtest runs after CASSANDRA-14420 + CASSANDRA-14421
* Introduce new node startup logfiles for stdout+stderr
* Use `./gradlew` instead of `ant` to build C*, if `gradlew` is present

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -711,14 +711,25 @@ def invalidate_cache():
     rmdirs(os.path.join(get_default_path(), 'repository'))
 
 
-def get_jdk_version():
+def get_jdk_version_int(process='java'):
+    jdk_version = float(get_jdk_version(process))
+    # Make it Java 8 instead of 1.8 (or 7 instead of 1.7)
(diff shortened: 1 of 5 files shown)


ccmlib/common.py
```diff
@@ -731,9 +742,118 @@ def _get_jdk_version(version):
...
+    # conservative Java version defaults
+    build_versions = [8]
+    run_versions = [8]
+
+    if not jvm_version:
+        if cassandra_version >= '4.0':
+            build_versions = [8, 11]
+            run_versions = [8, 11, 12, 13, 14, 15, 16, 17]
+        # Cassandra versions 3.x use the defaults
+        elif cassandra_version < '3.0':
```
nastra, line 796: seems like we don't need that `else` branch, since the values are the same as the defaults?

snazy:
Travis CI looks green: https://travis-ci.org/github/riptano/ccm/builds/694277941

PR #714, opened 2020-06-02
CASSANDRA-15835: Use the right Java version + JAVA_HOME when building C* from source / master

snazy:
Also:
* Add a new "version source type" `clone:/path/to/local/cassandra/clone` to re-enable upgrade-dtests for dev-CI and local upgrade-dtest runs after CASSANDRA-14420 + CASSANDRA-14421
* Introduce new node startup logfiles for stdout+stderr
* Use `./gradlew` instead of `ant` to build C*, if `gradlew` is present

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -703,15 +703,21 @@ def invalidate_cache():
     rmdirs(os.path.join(get_default_path(), 'repository'))
 
 
-def get_jdk_version():
+def get_jdk_version_int(process='java'):
+    jdk_version = float(get_jdk_version(process))
+    # Make it Java 8 instead of 1.8 (or 7 instead of 1.7)
+    jdk_version = int(jdk_version if jdk_version >= 2 else 10*(jdk_version-1))
+    return jdk_version
+
+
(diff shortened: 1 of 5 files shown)


ccmlib/common.py
```diff
@@ -728,9 +734,118 @@ def _get_jdk_version(version):
...
+    :return: the maybe updated OS environment variables. If 'env' was 'None' and JAVA_HOME needs to be set,
+    the result will also contain the current OS environment variables from 'os.environ'.
+    """
+
+    env = env if env else os.environ
+
+    current_java_version = get_jdk_version_int()
+    current_java_home_version = get_jdk_version_int('{}/bin/java'.format(env['JAVA_HOME'])) if 'JAVA_HOME' in env else current_java_version
+    # Code to ensure that we start C* using the correct Java version.
+    # This is important especially after DB-468 (Java 11 support) when dtests are run using Java 11
```
nastra, line 760: nit: should mention an OSS ticket?

snazy:
Travis CI looks green: https://travis-ci.org/github/riptano/ccm/builds/694280584

PR #718, opened 2020-08-26
Supporting changes for CASSANDRA-15937

jonmeredith:
Update address patterns for changes to InetAddressAn…dPort.toString

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -226,17 +226,17 @@ class Node(object):
 
     def address_for_current_version_slashy(self):
         """
-        Returns the address formatted for the current version with
-        the Java slashy if necessary.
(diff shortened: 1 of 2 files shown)


ccmlib/node.py
```diff
@@ -259,8 +259,12 @@ def address_and_port(self):
         """
         Returns the IP used for internal communication along with ports
         """
+        address = self.network_interfaces['storage'][0]
         port = self.network_interfaces['storage'][1]
-        return self.network_interfaces['storage'][0] + ':' + str(port)
+        if address.find(":") != -1:
+            return "[{}]:{}".format(address, port)
```
dcapwell, line 265: didn't test on ipv6 so not sure if something else will fail to handle the ipv6 address format...

PR #719, opened 2020-09-22
Add env. var for default start timeout

dimas-b:
This is mostly for Stargate PR https://github.com/stargate/stargate/pull/114

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -413,7 +413,9 @@ class Cluster(object):
             for node, p, mark in started:
                 try:
                     start_message = "Listening for thrift clients..." if self.cassandra_version() < "2.2" else "Starting listening for CQL clients"
-                    node.watch_log_for(start_message, timeout=kwargs.get('timeout',60), process=p, verbose=verbose, from_mark=mark)
+                    node.watch_log_for(start_message,
+                                       timeout=kwargs.get('timeout', int(os.environ.get('CCM_CLUSTER_START_DEFAULT_TIMEOUT', 60))),
+                                       process=p, verbose=verbose, from_mark=mark)
                 except RuntimeError:
                     return None
 


ccmlib/cluster.py
```diff
@@ -413,7 +413,7 @@ def start(self, no_wait=False, verbose=False, wait_for_binary_proto=True,
             for node, p, mark in started:
                 try:
                     start_message = "Listening for thrift clients..." if self.cassandra_version() < "2.2" else "Starting listening for CQL clients"
-                    node.watch_log_for(start_message, timeout=kwargs.get('timeout',60), process=p, verbose=verbose, from_mark=mark)
+                    node.watch_log_for(start_message, timeout=kwargs.get('timeout',600), process=p, verbose=verbose, from_mark=mark)
```
tlasica, line 416: I think one of the consequences is that if for some reason there was a problem with node startup then 
- before test failed after 1 min
- after test will fail after 10 min

I wonder what is the rationale behind making it that long?

dimas-b: Is time-to-failure-detection critical? I guess in most cases it is not since `ccm` is mostly used in test contexts (I guess).

The reason for the longer timeout is that in GH actions used to build Stargate the build nodes are slow and C* is not able to start in 1 min :man_shrugging:

dimas-b: As an alternative we can define a cmd. line parameter for this timeout, of course, but that would be a larger change, and I thought it was not really necessary. I'm open to reworking this PR.

tlasica: > Is time-to-failure-detection critical?
No, is not critical of course.
As it takes around 5 min to start (including download) then indeed 10 minutes makes sense...

tlasica (approved):
using env is probably less invasive for existing ccm usages

ccmlib/cluster.py
dimas-b: I've reworked this PR to use an env. var for the default timeout and force-pushed. Existing users should still get old behaviour.

nastra (approved):
LGTM. I wonder if this should eventually also be merged into the `cassandra-test` branch (which is being used by OSS C*)?

PR #720, opened 2020-10-02
Set metadata dir and add start timeout override

dimas-b:
This is intended to enable Stargate CI test with DSE 6.8 persistence: https://github.com/stargate/stargate/pull/182

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -412,10 +412,10 @@ class Cluster(object):
         else:
             for node, p, mark in started:
                 try:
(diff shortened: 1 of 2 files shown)


nastra (approved):
LGTM

tlasica (approved):
One thing is not super clear for me:
why do we need two env variables: for DEFAULT and for OVERRIDE?

dimas-b:
I added default last time for C*, but the DSE call path sets a hard-coded timeout of 180, hence the need to override as well.

PR #721, opened 2020-10-06
CASSANDRA-16127 - Addded new method to watch logs, but will fail if any errors are detected

dcapwell:


ccmlib/node.py
```diff
@@ -556,6 +559,40 @@ def watch_log_for(self, exprs, from_mark=None, timeout=600, process=None, verbos
...
+        This methods when all the expressions have been found or the method
+        timeouts (a TimeoutError is then raised). On successful completion,
+        a list of pair (line matched, match object) is returned.
+
+        This method is different from watch_log_for since it will fail if
+        any errors are found.
+        """
+        start = time.time()
+        tofind = [exprs] if isinstance(exprs, string_types) else exprs
+        tofind = [re.compile(e) for e in tofind]
```
dcapwell, line 574: this is only needed for the timeout message, so we can match `watch_log_for`

ccmlib/node.py
```diff
@@ -556,6 +559,40 @@ def watch_log_for(self, exprs, from_mark=None, timeout=600, process=None, verbos
                         if process.returncode == 0:
                             return None
 
+    def watch_log_for_no_errors(self, exprs, from_mark=None, timeout=600, process=None, verbose=False, filename='system.log'):
+        """
+        Watch the log until one or more (regular) expression are found.
```
ekaterinadimitrova2, line 564: expressionS :-)

ccmlib/node.py
```diff
@@ -556,6 +559,40 @@ def watch_log_for(self, exprs, from_mark=None, timeout=600, process=None, verbos
...
+        This methods when all the expressions have been found or the method
```
ekaterinadimitrova2, line 565: There is something wrong with this sentence :-) Another verb?

ccmlib/node.py
```diff
@@ -556,6 +559,40 @@ def watch_log_for(self, exprs, from_mark=None, timeout=600, process=None, verbos
...
+        timeouts (a TimeoutError is then raised). On successful completion,
+        a list of pair (line matched, match object) is returned.
+
+        This method is different from watch_log_for since it will fail if
```
ekaterinadimitrova2, line 569: fail Cassandra

ekaterinadimitrova2 (commented):
Looks good to me as soon as no tests got broken :-) But I doubt it :-)

ccmlib/node.py
dcapwell: I copy/pasted from watch_log_for and didn't read =D

ccmlib/node.py
dcapwell: ill rewrite and fix watch_log_for

ccmlib/node.py
dcapwell: we don't stop cassandra, we only stop watching the logs.

ekaterinadimitrova2: True, my point is this fail doesn't tell me what failure we talk about :-)

ccmlib/node.py
ekaterinadimitrova2: I didn't check the other one. :-)

ccmlib/node.py
ekaterinadimitrova2: Thanks :-)

ccmlib/node.py
dcapwell: How about

> This method is different from watch_log_for as it will raise a AssertionError if the logs contain an error; this assertion will contain the errors found in the logs.

PR #724, opened 2021-01-14
STAR-57. Check if pid is alive while waiting for logs

tlasica:
Rationale:
When starting C* ccm checks logs, waiting for particular phrases.
We want to fail test asap if there is a problem with node start.

Before:
If there is any problem with node start it will not happen and start() will hit timeout, by default 90s.
So the test will fail after 90s while it actually fail after 5s.

After:
If logs are streamed ccm will check if node pid is still alive.
If not it will immediately raise `NodeError` to fail test telling that node is not started correctly.

Note: 
this may break some tests, that are testing node start failure by expecting `TimeoutError`.
Such tests should be adjusted, but there should be only a few.

8ba2709f9c4080f18229513d0b9d8bd028bd98c3 is about some refactoring
8a960d9b6473d941d3dc687bee04c3a6ac133f9e is the change

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -8,14 +8,13 @@ import re
 import shutil
 import signal
 import subprocess
-import sys
 import threading
 import time
 from collections import OrderedDict, defaultdict, namedtuple
 from distutils.version import LooseVersion #pylint: disable=import-error, no-name-in-module
 
 import yaml
-from six import iteritems, print_
+from six import print_
 
 from ccmlib import common, extension, repository
 from ccmlib.node import Node, NodeError, TimeoutError
@@ -264,11 +263,7 @@ class Cluster(object):
         dcs = []
 
         if use_vnodes is None:
-            self.use_vnodes = (
-                (tokens is not None and len(tokens) > 1)
-                    or ('num_tokens' in self._config_options
-                        and self._config_options['num_tokens'] is not None
-                        and int(self._config_options['num_tokens']) > 1))
+            self.use_vnodes = len(tokens or []) > 1 or self._more_than_one_token_configured()
         else:
             self.use_vnodes = use_vnodes
 
@@ -374,11 +369,13 @@ class Cluster(object):
         tokens.extend(new_tokens)
         return tokens
 
+    def _more_than_one_token_configured(self):
+        return int(self._config_options.get('num_tokens', '1')) > 1
+
     def can_generate_tokens(self):
         return (self.cassandra_version() >= '4'
                     and (self.partitioner is None or ('Murmur3' in self.partitioner or 'Random' in self.partitioner))
-                    and ('num_tokens' in self._config_options
-                            and self._config_options['num_tokens'] is not None and int(self._config_options['num_tokens']) > 1))
+                    and self._more_than_one_token_configured)
 
     def generated_tokens(self, dcs):
         tokens = []
@@ -398,7 +395,7 @@ class Cluster(object):
     def generate_dc_tokens(self, node_count, tokens):
         if self.cassandra_version() < '4' or (self.partitioner and not ('Murmur3' in self.partitioner or 'Random' in self.partitioner)):
             raise common.ArgumentError("generate-tokens script only for >=4.0 and Murmur3 or Random")
-        if not ('num_tokens' in self._config_options and self._config_options['num_tokens'] is not None and int(self._config_options['num_tokens']) > 1):
+        if not self._more_than_one_token_configured():
             raise common.ArgumentError("Cannot use generate-tokens script without num_tokens > 1")
 
         partitioner = 'RandomPartitioner' if ( self.partitioner and 'Random' in self.partitioner) else 'Murmur3Partitioner'
@@ -518,7 +515,6 @@ class Cluster(object):
 
                 # if the node is going to allocate_strategy_ tokens during start, then wait_for_binary_proto=True
                 node_wait_for_binary_proto = (self.can_generate_tokens() and self.use_vnodes and node.initial_token is None)
-
                 p = node.start(update_pid=False, jvm_args=jvm_args, jvm_version=jvm_version,
                                profile_options=profile_options, verbose=verbose, quiet_start=quiet_start,
                                allow_root=allow_root, wait_for_binary_proto=node_wait_for_binary_proto)
@@ -536,11 +532,15 @@ class Cluster(object):
             time.sleep(2)  # waiting 2 seconds to check for early errors and for the pid to be set
         else:
             for node, p, mark in started:
+                if not node._wait_for_running(p, timeout_s=7):
+                    raise NodeError("Node {} should be running before waiting for <started listening> log message, "
+                                    "but C* process is terminated.".format(node.name))
                 try:
                     timeout=kwargs.get('timeout', DEFAULT_CLUSTER_WAIT_TIMEOUT_IN_SECS)
                     timeout=int(os.environ.get('CCM_CLUSTER_START_TIMEOUT_OVERRIDE', timeout))
                     start_message = "Listening for thrift clients..." if self.cassandra_version() < "2.2" else "Starting listening for CQL clients"
-                    node.watch_log_for(start_message, timeout=timeout, process=p, verbose=verbose, from_mark=mark)
+                    node.watch_log_for(start_message, timeout=timeout, process=p, verbose=verbose, from_mark=mark,
+                                       error_on_pid_terminated=True)
                 except RuntimeError:
                     return None
 
@@ -821,11 +821,11 @@ class Cluster(object):
         Returns the first node where the pattern was found, along with the matching lines.
         Raises a TimeoutError if the pattern is not found within the specified timeout period.
         """
-        end_time = time.time() + timeout_seconds
+        start_time = time.time()
         while True:
-            if time.time() > end_time:
-                raise TimeoutError(time.strftime("%d %b %Y %H:%M:%S", time.gmtime()) +
-                                   " Unable to find: " + versions_to_patterns.patterns + " in any node log within " + str(timeout_seconds) + "s")
+            TimeoutError.raise_if_passed(start=start_time, timeout=timeout_seconds,
+                                         msg="Unable to find: {x} in any node log within {t} s".format(
+                                             x=versions_to_patterns.patterns, t=timeout_seconds))
 
             for node in self.nodelist():
                 pattern = versions_to_patterns(node.get_cassandra_version())
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -19,6 +19,7 @@ import sys
 import time
 from distutils.version import LooseVersion  #pylint: disable=import-error, no-name-in-module
 
+import six
 import yaml
 from six import print_
 
@@ -452,12 +453,7 @@ def get_stress_bin(install_dir):
         add_exec_permission(path, short_bin)
     elif not os.access(stress, os.X_OK):
         try:
-            # try to add user execute permissions
-            # os.chmod doesn't work on Windows and isn't necessary unless in cygwin...
-            if sys.platform == "cygwin":
-                add_exec_permission(path, stress)
-            else:
-                os.chmod(stress, os.stat(stress).st_mode | stat.S_IXUSR)
+            os.chmod(stress, os.stat(stress).st_mode | stat.S_IXUSR)
         except:
             raise Exception("stress binary is not executable: %s" % (stress,))
 
@@ -546,7 +542,6 @@ def check_socket_listening(itf, timeout=60):
                 sock.close()
             # Try again in another 200ms
             time.sleep(.2)
-            continue
 
     return False
 
@@ -885,13 +880,12 @@ def merge_configuration(original, changes, delete_empty=True, delete_always=Fals
     return new
 
 
+def is_int_not_bool(obj):
+    return isinstance(obj, six.integer_types) and not isinstance(obj, bool)
+
+
 def is_intlike(obj):
-    try:
-        int(obj)
-        return True
-    except TypeError:
-        return False
-    raise RuntimeError('Reached end of {}; should not be possible'.format(is_intlike.__name__))
+    return isinstance(obj, six.integer_types)
 
 
 def wait_for_any_log(nodes, pattern, timeout, filename='system.log', marks=None):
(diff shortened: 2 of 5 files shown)


tlasica:
Kind of interesting that build fails on python 3.7.7

ccmlib/node.py
```diff
@@ -551,6 +551,12 @@ def watch_log_for(self, exprs, from_mark=None, timeout=600, process=None, verbos
                     if start + timeout < time.time():
                         raise TimeoutError(time.strftime("%d %b %Y %H:%M:%S", time.gmtime()) + " [" + self.name + "] Missing: " + str([e.pattern for e in tofind]) + ":\n" + reads[:50] + ".....\nSee {} for remainder".format(filename))
 
+                    # update status because meanwhile node could have died....
+                    pid_before_check = self.pid
+                    self.__update_status()
+                    if not self.is_running():
+                        raise RuntimeError("node {n} with pid {pid} is not running".format(n=self.name, pid=pid_before_check))
```
tlasica, line 558: I wonder if it can cause problems... in the `node.stop()`
it should not but there is a possible race between reading log and the moment pid disappears

tlasica:
retest this please

tlasica:
hi, may I ask why did you close this PR @michaelsembwever  ?

michaelsembwever:
That was unintentional. 

It looks like when I deleted the `cassandra-test` branch all PRs pointing to it got automatically closed.

ref: https://issues.apache.org/jira/browse/CASSANDRA-16383

tlasica:
@michaelsembwever I see, so I can just rebase branch to master ...

michaelsembwever:
> I can just rebase branch to master ...

Yes :)

ccmlib/cluster.py
```diff
@@ -518,8 +516,7 @@ def start(self, no_wait=False, verbose=False, wait_for_binary_proto=True,
 
                 # if the node is going to allocate_strategy_ tokens during start, then wait_for_binary_proto=True
                 node_wait_for_binary_proto = (self.can_generate_tokens() and self.use_vnodes and node.initial_token is None)
-
-                p = node.start(update_pid=False, jvm_args=jvm_args, jvm_version=jvm_version,
+                p = node.start(update_pid=True, jvm_args=jvm_args, jvm_version=jvm_version,
```
tlasica, line 519: this is tricky change... I am not 100% it is correct 
I wonder why we could start with `update_pid=False`?

ccmlib/common.py
```diff
@@ -452,12 +453,7 @@ def get_stress_bin(install_dir):
         add_exec_permission(path, short_bin)
     elif not os.access(stress, os.X_OK):
         try:
-            # try to add user execute permissions
-            # os.chmod doesn't work on Windows and isn't necessary unless in cygwin...
-            if sys.platform == "cygwin":
-                add_exec_permission(path, stress)
-            else:
-                os.chmod(stress, os.stat(stress).st_mode | stat.S_IXUSR)
+            os.chmod(stress, os.stat(stress).st_mode | stat.S_IXUSR)
```
tlasica, line 456: please notice that `if sys.platform == cygwin` is branched one level earlier, so we cannot get to this place

ccmlib/node.py
```diff
@@ -558,21 +558,35 @@ def watch_log_for(self, exprs, from_mark=None, timeout=600, process=None, verbos
                             matchings.append((line, m))
                             tofind.remove(e)
                             if len(tofind) == 0:
+                                common.debug("return")
```
tlasica, line 561: TO BE REMOVED

ccmlib/node.py
```diff
@@ -558,21 +558,35 @@ def watch_log_for(self, exprs, from_mark=None, timeout=600, process=None, verbos
...
-                            return None
+                    # update status because meanwhile node could have died....
+                    # if it was running but now it is not we should return immediately rather than after timeout
+                    if self.pid and not self._is_pid_running():
+                        msg = "node {n} with pid {pid} stopped during watching log for {m}".\
+                            format(n=self.name, pid=self.pid, m=exprs)
+                        common.warning(msg)
+                        raise NodeError(msg)
+
+                # process is in this case process that started C*
```
tlasica, line 577: This is another tricky part....

Possibly the intention of this part of the code was to actually finish immediately if child process terminated... 
`return None` as not supper effective as results are not checked.

There is one interesting thing:
when checking `None` it can fail positive tests: `test3` or `test2` which is kind of unexpected.

ccmlib/node.py
```diff
@@ -639,7 +653,14 @@ def wait_for_binary_interface(self, **kwargs):
         Emits a warning if not listening after NODE_WAIT_TIMEOUT_IN_SECS seconds.
         """
         if self.cluster.version() >= '1.2':
-            self.watch_log_for("Starting listening for CQL clients", **kwargs)
+            start_time = time.time()
```
tlasica, line 656: TODO: remove logging

ccmlib/node.py
```diff
@@ -1857,55 +1878,57 @@ def _update_topology_file(self):
         with open(topology_file, 'w') as f:
             f.write(content)
 
+    def _is_pid_running(self):
```
tlasica, line 1881: this is refactoring to make status updating logic simpler

ccmlib/node.py
```diff
@@ -565,14 +565,26 @@ def watch_log_for(self, exprs, from_mark=None, timeout=600, process=None, verbos
...
-                            return None
+                    # update status because meanwhile node could have died....
+                    # if it was running but now it is not we should return immediately rather than after timeout
+                    if self.pid and not self._is_pid_running():
+                        msg = "node {n} with pid {pid} stopped during watching log for {m}".\
+                            format(n=self.name, pid=self.pid, m=exprs)
+                        common.warning(msg)
+                        raise NodeError(msg)
+
+                # this can preliminary end checking (if child process is terminated)
```
tlasica, line 576: IIUC: if child process terminated => return None
but this None is not checked anywhere....

ccmlib/cluster.py
michaelsembwever: 🤷

michaelsembwever (commented):
Changes LGTM. (thanks for the clean-up and the cleaner code alternatives).

I'm keen to see a full ci-cassandra run against this, when ready.
And what the heck `update_pid` does…?

tlasica:
@michaelsembwever  thanks!

the fix is incomplete or simply incorrect... I have run full cassandra-dtest suite and there are some problems, for example with tests that by design want to crash nodes on CRC problems. so it will have to be more subtle.

ccmlib/node.py
```diff
@@ -630,17 +659,30 @@ def watch_log_for_alive(self, nodes, from_mark=None, timeout=120, filename='syst
         tofind = ["%s.* now UP" % node.address_for_version(self.get_cassandra_version()) for node in tofind]
         self.watch_log_for(tofind, from_mark=from_mark, timeout=timeout, filename=filename)
 
+    def raise_node_error_if_cassandra_process_is_terminated(self):
+        up = self._is_pid_running()
+        common.debug("checking if C* process with pid {pid} is running: {up}".format(pid=self.pid, up=up))
```
tlasica, line 664: to be removed after testing

tlasica:
@michaelsembwever  This one is ready for re-review. 

Would be great to have 2nd pair of eyes.

tlasica:
Running some circleCI: https://app.circleci.com/pipelines/github/tlasica/cassandra?branch=ccm-star-57

There are two tests that need to catch both TimeoutError and NodeError (as technically both can be thrown).
In fact even before that change both can be thrown.

tlasica:
both runs [some test adjustments I plan to add as CASSANDRA-14605] (https://github.com/apache/cassandra-dtest/compare/trunk...tlasica:ccm-star-57)  look pretty good to me
https://ci-cassandra.apache.org/view/patches/job/Cassandra-devbranch/332/ against trunk
https://ci-cassandra.apache.org/view/patches/job/Cassandra-devbranch/330/ against cassandra-3.11

ccmlib/cluster.py
```diff
@@ -374,11 +369,14 @@ def balanced_tokens_across_dcs(self, dcs):
         tokens.extend(new_tokens)
         return tokens
 
+    def _more_than_one_token_configured(self):
+        num_tokens = self._config_options.get('num_tokens', None)
```
jacek-lewandowski, line 373: maybe just `int(self._config_options.get('num_tokens', '1')) > 1`

ccmlib/cluster.py
```diff
@@ -536,11 +533,15 @@ def start(self, no_wait=False, verbose=False, wait_for_binary_proto=True,
             time.sleep(2)  # waiting 2 seconds to check for early errors and for the pid to be set
         else:
             for node, p, mark in started:
+                if not node._wait_for_running(p, timeout_s=7):
```
jacek-lewandowski, line 536: is if we have say 3 nodes, we will wait up to 7 seconds for each in a sequence?

ccmlib/node.py
```diff
@@ -519,12 +528,19 @@ def watch_log_for(self, exprs, from_mark=None, timeout=600, process=None, verbos
         if len(tofind) == 0:
             return None
 
+        def check_timeout(msg):
+            if start + timeout < time.time():
+                tstamp = time.strftime("%d %b %Y %H:%M:%S", time.gmtime())
```
jacek-lewandowski, line 533: nit: time format could be extracted to some global / static / final field?

jacek-lewandowski: or maybe this whole function deserves to be top level - it looks pretty generic

ccmlib/node.py
```diff
@@ -560,19 +576,32 @@ def watch_log_for(self, exprs, from_mark=None, timeout=600, process=None, verbos
...
-                if process:
-                    if common.is_win():
-                        if not self.is_running():
-                            return None
-                    else:
-                        process.poll()
-                        if process.returncode == 0:
-                            return None
+                    # check if abort condition is not met e.g. C* process is terminated
+                    if abort_function is not None:
```
jacek-lewandowski, line 583: why do we call it here?

ccmlib/node.py
```diff
@@ -560,19 +576,32 @@ def watch_log_for(self, exprs, from_mark=None, timeout=600, process=None, verbos
...
+                    # check if timeout passed
+                    check_timeout("Missing: " + str([e.pattern for e in tofind]) +
+                                  ":\n" + reads[:50] +
+                                  ".....\nSee {} for remainder".format(filename))
+
+                    # Checking "process" is tricky, as it may be itself terminated e.g. after "verbose"
+                    # or if there is some race condition between log checking and start process finish
+                    # I am simply afraid to completely remove it because this is general usage method
+                    # so instead we will use it IFF the pid-based abort function is not provided
+                    if process and abort_function is None:
```
jacek-lewandowski, line 595: isn't `abort_function is None` redundant?

ccmlib/node.py
jacek-lewandowski: ok, the logic of abort function mixed me up - by calling `abort_function` I conclude that we want to abort at this point. But it is more like maybe_abort... can be this more explicit? Maybe just pass a bool param - can_abort or something and then inline that cancellation code?

ccmlib/node.py
```diff
@@ -836,30 +879,37 @@ def start(self,
...
                 node.watch_log_for_alive(self, from_mark=mark, timeout=wait_other_notice)
         elif wait_other_notice:
             for node, mark in marks:
                 node.watch_log_for_alive(self, from_mark=mark)
 
-        # If wait_for_binary_proto is a bool, we don't want to treat it as a
-        # timeout. Other intlike types, though, we want to use.
-        if common.is_intlike(wait_for_binary_proto) and not isinstance(wait_for_binary_proto, bool):
+        # if requested wait for binary protocol to start
+        if common.is_int_not_bool(wait_for_binary_proto):
```
jacek-lewandowski, line 897: what about composing two variables 

```python
should_wait_for_binary_proto = wait_for_binary_proto is not None
binary_proto_timeout = wait_for_binary_proto if is_int_not_bool(wait_for_binary_proto) else None
``` 

and then just call 

```python
if should_wait_for_binary_proto: 
    self.wait_for_binary_interface(from_mark=self.mark, timeout=binary_proto_timeout)
```

jacek-lewandowski (commented):
what about https://github.com/riptano/ccm/pull/724/files#diff-07cfe037dd1d2ac633d6a2ee26160a2dc94bc7bb8ba4efe7c329cbf2192d8dbfR565 when the process exits with 0?

ccmlib/cluster.py
tlasica: yes, but it does not matter...
because if they are started in parallel... then after we catch 1st running we will with high probability immediately have others running as well.

ccmlib/node.py
tlasica: I was thinking about it as well... but the problem is this `watch_for_logs` is generic, used in tons of places.
So I wanted to have control about the condition that is used to abort...

I can try to change it again and instead of a function use a boolean: `stop_if_cassandra_pid_terminated`...

ccmlib/node.py
tlasica: Not sure why redundant?
It is indeed checked earlier BUT it cannot not throw (it is conditional).
Indeed, I need to change naming ;-)


this is what makes this whole change so complex:
process - is a process that started C*
abort function is a function based e.g. on C* process pid (different one).

Now there are interesting races:
this process can be terminated before C* process is terminated.
in such case instead of finding C* process dead => raise NodeError
this logic will return None and continue
and then it will timeout on the next step or fail silently

ccmlib/node.py
tlasica: this will not work straight because `wait_for_binary_proto` can be `False`.

ccmlib/node.py
jacek-lewandowski: or just pass a condition

ccmlib/node.py
tlasica: I would rather not touch it...

jacek-lewandowski (approved):

tlasica:
Running new round of tests after recent changes:
https://ci-cassandra.apache.org/view/patches/job/Cassandra-devbranch/338/ cassandra-3.11
https://ci-cassandra.apache.org/view/patches/job/Cassandra-devbranch/339/ trunk

michaelsembwever (approved):

michaelsembwever:
merged with https://github.com/riptano/ccm/commit/89af9303b9edad55f02883170fc1e100065c1445

PR #727, opened 2021-01-29
CASSANDRA-16411 CASSANDRA-16296 Fix topology corruption on joining no…

bereng:
…des without DC

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,6 +261,15 @@ The completion script relies on ccm having two hidden subcommands:
 
 Thus, it will not work with sufficiently old versions of ccm.
 
+Testing
+-----------------------
+
+Create a virtual environment i.e.:
+
+    python3 -m venv ccm
+
+Pip install all dependecies and `mock`. You'll also need `nose` if `nosetests` is not installed in your system already. Now cd into the `tests` folder and run `nosetests`.
+
 Remote debugging
 -----------------------
 
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -240,6 +240,11 @@ class Cluster(object):
             self.seeds.append(node)
         self._update_config()
         node.data_center = data_center
+        if data_center is None:
+            for existing_node in self.nodelist():
+                if existing_node.data_center is not None:
+                    raise common.ArgumentError('Please specify the DC this node should be added to')
+
         node.set_log_level(self.__log_level)
 
         for debug_class in self._debug:
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -247,8 +247,16 @@ class Cluster(object):
         for trace_class in self._trace:
             node.set_log_level("TRACE", trace_class)
 
+        if data_center is None:
+            for existingNode in self.nodelist():
+                if existingNode.data_center is not None:
+                    node.data_center = "dc1"
+                    data_center = "dc1"
+                    break;
+
         if data_center is not None:
             self.__update_topology_files()
+
         node._save()
         return self
 
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -252,13 +252,6 @@ class Cluster(object):
         for trace_class in self._trace:
             node.set_log_level("TRACE", trace_class)
 
-        if data_center is None:
-            for existingNode in self.nodelist():
-                if existingNode.data_center is not None:
-                    node.data_center = "dc1"
-                    data_center = "dc1"
-                    break;
-
         if data_center is not None:
             self.__update_topology_files()
 
--- a/tests/test_lib.py
+++ b/tests/test_lib.py
@@ -292,15 +292,8 @@ class TestCCMLib(ccmtest.Tester):
                                          remote_debug_port='0',
                                          initial_token=None,
                                          binary_interface=('127.0.0.3', 9042))
-        try:
+        with self.assertRaisesRegexp(ccmlib.common.ArgumentError, 'Please specify the DC this node should be added to'):
             self.cluster.add(node3, is_seed=False)
-            self.fail("Should have requested a mandatory DC when adding the node.")
-        except ccmlib.common.ArgumentError:
-            pass
-        finally:
-            self.cluster.cleanup()
-            self.cluster.clear()
-            self.cluster.stop()
 
 class TestRunCqlsh(ccmtest.Tester):
 
--- a/tests/test_lib.py
+++ b/tests/test_lib.py
@@ -280,6 +280,28 @@ class TestCCMLib(ccmtest.Tester):
             self.cluster.clear()
             self.cluster.stop()
 
+    def test_dc_mandatory_on_multidc(self):
+        self.cluster = Cluster(CLUSTER_PATH, "mandatorydc", cassandra_version='git:trunk')
+        self.cluster.populate([1, 1])
+
+        node3 = self.cluster.create_node(name='node3',
+                                         auto_bootstrap=True,
+                                         thrift_interface=('127.0.0.3', 9160),
+                                         storage_interface=('127.0.0.3', 7000),
+                                         jmx_port='7300',
+                                         remote_debug_port='0',
+                                         initial_token=None,
+                                         binary_interface=('127.0.0.3', 9042))
+        try:
+            self.cluster.add(node3, is_seed=False)
+            self.fail("Should have requested a mandatory DC when adding the node.")
+        except ccmlib.common.ArgumentError:
+            pass
+        finally:
+            self.cluster.cleanup()
+            self.cluster.clear()
+            self.cluster.stop()
+
 class TestRunCqlsh(ccmtest.Tester):
 
     def setUp(self):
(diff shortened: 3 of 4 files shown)


ccmlib/cluster.py
```diff
@@ -247,8 +247,16 @@ def add(self, node, is_seed, data_center=None):
         for trace_class in self._trace:
             node.set_log_level("TRACE", trace_class)
 
+        if data_center is None:
```
tlasica, line 250: if I understand correctly the logic here is:
"if data center is not set for then newly added node should get DC from first DC used in existing cluster nodes"

so we can move L242 after calculation of initial `data_center`
```
if data_center is None:
    cluster_dcs = [n.data_center for n in self.nodelist() if n.data_center is not None]
    data_center = cluster_dcs[0] if cluster_dcs else None
node.data_center = data_center
```

tlasica: I wonder if it is possible that some test / usage would like to use node without DC being set?
and in such case using `None` will not work, as it will be ignored.

README.md
```diff
@@ -261,6 +261,15 @@ The completion script relies on ccm having two hidden subcommands:
...
 Thus, it will not work with sufficiently old versions of ccm.
 
+Testing
+-----------------------
+
+Create a virtual environment i.e.:
+
+    python3 -m venv ccm
+
+Pip install all dependecies and `mock`. You'll also need `nose` if `nosetests` is not installed in your system already. Now cd into the `tests` folder and run `nosetests`.
```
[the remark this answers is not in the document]
bereng: It didn't for me, but I am no `nosetests` expert :shrug: If sbdy can confirm either way I'll be happy to amend.

ccmlib/cluster.py
tlasica: I have another doubt:
if there are different DCs used in the cluster, and someone adds... it will get first DC.

so maybe instead of quessing / overriding we should assert that DC is provided if there is more than 1 dc in the cluster?

In such case we can assert `if data_center is None:`
```
assert len(set(cluster_dcs)) <= 1, "..." 
```

README.md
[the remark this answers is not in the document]
tlasica: ok

ccmlib/cluster.py
bereng: DC can be None when there are no DC's defined and it is fine (`SimpleStrategy`). Everybody jumps onto the 'default' one.

On NTPS a DC of None will have the node join the default DC so we need to account for that. I went for the default one when `None` instead of relying on the list of nodes as I don't have ordering guarantees afaik on the list.

I don't think a multi-DC setup with respecting 'None' makes sense as it breaks topology which is exactly what we are fixing here. Hope it makes sense.

ccmlib/cluster.py
```diff
@@ -247,8 +247,16 @@ def add(self, node, is_seed, data_center=None):
...
+            for existingNode in self.nodelist():
```
tlasica, line 251: naming should be `existing_node` to not break flake8 rules.

ccmlib/cluster.py
tlasica: I have updated the proposal to reflect the fact that we set to "dc1" explicitly,
assuming that any new node will join first dc
and first dc in ccm is dc1

tlasica: if we want to reflect "always use dc1 if any dc is used" then
```
if data_center is None:
    cluster_dcs = [n.data_center for n in self.nodelist() if n.data_center is not None]
    data_center = "dc1" if cluster_dcs else None
node.data_center = data_center
```

ccmlib/cluster.py
bereng: ha! good catch. My java brain kicking in lol...

ccmlib/cluster.py
bereng: Erroring out on missing DC when there were multiple DCs present was my first proposal. But then I noticed ccm forces dc1 and that is the default one as well. So I assumed it safe to use dc1 directly.

Some form of `len(nodelist().data_center) !=0` should work to the same effect. Anyway I say we wait for further reviews before pushing changes to bulk change in one go.

bereng: [Here](https://app.circleci.com/pipelines/github/bereng/cassandra/199/workflows/dbcb0f6e-1bb0-4334-b60c-07f07cb71792/jobs/1741) is the run where I would fail if no dc was specified. There were like 20-ish failures but it would be easy to fix those if we prefer that option.

tlasica: I think it is better to explicitly say in a multidc scenario which dc the node is added to
https://www.python.org/dev/peps/pep-0020/#id2

ekaterinadimitrova2: @bereng, I saw only three tests failing, is this the right link? 
My personal preference - do it as close as possible to the reality and any implications or possible second guesses/confusion - commented + warnings logged so that people use this correctly and have the right expectations

ccmlib/cluster.py
```diff
@@ -247,8 +247,16 @@ def add(self, node, is_seed, data_center=None):
...
+                if existingNode.data_center is not None:
+                    node.data_center = "dc1"
+                    data_center = "dc1"
+                    break;
```
ekaterinadimitrova2, line 255: remove the ";"

ccmlib/cluster.py
bereng: Ok switching to erroring out if DC is missing.

bereng:
Swtiched to erroring out, added a ccm test and on the dtest PR you can see the dtests changes + CI runs.

bereng:
Apologies guys, I had left behind some code from the previous commit, my bad. Removed now.

tests/test_lib.py
```diff
@@ -280,6 +280,28 @@ def test_node_start_with_non_default_timeout(self):
...
+
+        node3 = self.cluster.create_node(name='node3',
+                                         auto_bootstrap=True,
+                                         thrift_interface=('127.0.0.3', 9160),
+                                         storage_interface=('127.0.0.3', 7000),
+                                         jmx_port='7300',
+                                         remote_debug_port='0',
+                                         initial_token=None,
+                                         binary_interface=('127.0.0.3', 9042))
+        try:
```
tlasica, line 295: nit: in `pytest` instead of `try..except` we can use:
```
with pytest.raises(ArgumentError):
    self.cluster.add(node3, is_seed=False)
```
https://docs.pytest.org/en/reorganize-docs/new-docs/user/pytest_raises.html

cluster will be cleaned up by Tester teardown.

tlasica (approved):

ccmlib/cluster.py
```diff
@@ -240,15 +240,28 @@ def add(self, node, is_seed, data_center=None):
...
+                    raise common.ArgumentError('Please specify the DC this node should be added to')
+
         node.set_log_level(self.__log_level)
 
         for debug_class in self._debug:
             node.set_log_level("DEBUG", debug_class)
         for trace_class in self._trace:
             node.set_log_level("TRACE", trace_class)
 
+        if data_center is None:
```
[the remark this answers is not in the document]
bereng: Mmmm this comment seems gone now. When `None` topology file is not updated, hence the default `dc1` hardcoded in that file will be picked up. Does it answer the q?

bereng:
I have added CI for this latest commit on the dtests PR to keep everything in the same place #justfyi

ccmlib/cluster.py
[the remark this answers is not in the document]
tlasica: yes, it is a leftover from the copy-paste wraith ;-)

tests/test_lib.py
```diff
@@ -280,6 +280,28 @@ def test_node_start_with_non_default_timeout(self):
...
+            self.cluster.add(node3, is_seed=False)
+            self.fail("Should have requested a mandatory DC when adding the node.")
+        except ccmlib.common.ArgumentError:
+            pass
```
adelapena, line 299: Nit: Maybe we can check the exception message?
```suggestion
            self.assertEqual('Please specify the DC this node should be added to', str(e))
            pass
```

adelapena (approved):
Changes look good to me, even though I'm not very familiarized with ccm

tests/test_lib.py
tlasica: if we go this path then maybe more pythonic (should work)```
with self.assertRaisesRegexp(cccmlib.common.ArgumentError, 'Please specify (...)'):
    self.cluster.add(node3, is_seed=False)
```

**Review by ekaterinadimitrova2 — COMMENTED** [review]
I am also not one of the ccm code gurus on the project but I think the proposed solution is deterministic and makes the things clear. I would run it also as a comment in the slack channel to see if it might affect someone in a way but I don't see how. My only concern is it affects the CI before a release :-) Good catch of the bug

File: ccmlib/cluster.py [hunk 8]
```diff
@@ -240,15 +240,28 @@ def add(self, node, is_seed, data_center=None):
             self.seeds.append(node)
         self._update_config()
         node.data_center = data_center
+        if data_center is None:
```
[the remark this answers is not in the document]
bereng: Being a nit and single case I am choosing not to extract it as a method :man_shrugging:. Personal preference

bereng:
I can run it through Slack +1

ekaterinadimitrova2 (approved):

PR #735, opened 2021-12-16
ci: add py39, py310

chenrui333:
- relates to https://github.com/Homebrew/homebrew-core/pull/86552


chenrui333:
looks like no build history in the past several months
image

PR #736, opened 2022-01-20
Helper script to provision a CCM cluster in MacOS

frankgh:
The ccm-macos.bash script allows the creation of a CCM cluster. It supports the `NUM_NODES`
parameter to determine the number of nodes to configure in the cluster. Additionally, it will
configure the required interfaces required for the cluster to start up. It requires `sudo` for
the operation, so this script will need elevated privileges to configure the interfaces
required.

Additionally, this script relies on a configured virtualenv. Conveniently, it can leverage
Cassandra's pylib virtualenv. If the virtualenv is not loaded, it will check whether
Cassandra's pylib virtualenv has been already initialized. If so, then it will activate the
virtualenv. Otherwise, it will initialize the virtualenv and install the requirements.

The name of the cluster defaults to `test`, but it can be configured by providing the
`CLUSTER_NAME` parameter.

### DIFF
--- a/misc/ccm-macos.bash
+++ b/misc/ccm-macos.bash
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+CLUSTER_NAME=${CLUSTER_NAME:-test}
+NUM_NODES=${NUM_NODES:-3}
+CASSANDRA_DIR=${CASSANDRA_DIR:-$HOME/cassandra}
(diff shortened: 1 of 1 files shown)


frankgh:
@driftx would you be able to take a look at this?

driftx:
I don't have a mac to test with but this looks good to me.

PR #737, opened 2022-02-15
Update Homebrew link in README.md

justinchuch:
brew link is outdated, updated accordingly. Thanks @thobe for reporting.


thobe (approved):
This change looks like it does exactly what it should do!

PR #744, opened 2022-09-21
CASSANDRA-17861 Updated README.md to reference unittest instead of nose

bschoening:


README.md
```diff
@@ -264,11 +263,13 @@ Thus, it will not work with sufficiently old versions of ccm.
...
 -----------------------
 
+Tests are written with the standard library _unittest_ package.
+
 Create a virtual environment i.e.:
 
     python3 -m venv ccm
 
-Pip install all dependecies and `mock`. You'll also need `nose` if `nosetests` is not installed in your system already. Now cd into the `tests` folder and run `nosetests`.
+Pip install all dependecies and `mock`. Now cd into the `tests` folder and run `python -m unittest`.
```
ekaterinadimitrova2, line 272: I was thinking maybe we can add more detailed commands here - similar to how we do for dtests - 
https://github.com/apache/cassandra-dtest#python-dependencies
That way even someone not very familiar with python, etc will be able to do a quick start, WDYT?

ekaterinadimitrova2: Also, the tests really fail with some imports at the moment :( 
Side note: ccm needs to be compatible to Python 2 and 3 as far as I know

bschoening: I see why dtest might have that because it's testing a Java project.  But if you're changing ccm and wish to run the unit tests, you're probably a python developer.

ekaterinadimitrova2: I agree with you, but I'd say we try to be inclusive and consider also Python developers in development who might be learning and testing hehe

PR #746, opened 2022-11-11
Fix broken nosetest to pytest fixtures

aweisberg:
Also update README.md with a bunch of corrected/new info.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,17 +7,52 @@ localhost.
 The goal of ccm and ccmlib is to make it easy to create, manage and destroy a
 small Cassandra cluster on a local box. It is meant for testing a Cassandra cluster.
 
+New to Python development?
+--------------------------
+Python has moved on since CCM started development. `pip` is the new `easy_install`,
+Python 3 is the new 2.7, and pyenv and virtualenv are strongly recommended for managing
+multiple Python versions and dependencies for specific Python applications.
+
+A typical MacOS setup would be to install [Homebrew](https://docs.brew.sh/Installation),
+then `brew install pyenv` to manage Python versions and then use virtualenv to
+manage the dependencies for CCM. Make sure to add [brew's bin directory to your path in
+your ~/.zshenv](https://www.zerotohero.dev/zshell-startup-files/). This would be
+`/usr/local` for MacOS Intel and `/opt/homebrew/` for MacOS on Apple Silicon.
+
+Now you are ready to install Python using pyenv. To avoid getting a bleeding edge version that will fail with 
+some aspect of CCM you can `pyenv install 3.9.16`.
+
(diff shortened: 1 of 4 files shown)


bereng:
You may want to mention pyenv as a hint in cases people need to have multiple python installations on the box

README.md
```diff
@@ -7,17 +7,47 @@ localhost.
...
 - A working python installation (tested to work with python 2.7).
-- pyYAML (http://pyyaml.org/ -- `sudo easy_install pyYaml`)
-- six (https://pypi.org/project/six/ -- `sudo easy_install six`)
+- See `requirements.txt` for runtime requirements
+- `mock` and `pytest` for tests
 - ant (http://ant.apache.org/, on Mac OS X, `brew install ant`)
-- psutil (https://pypi.org/project/psutil/)
 - Java (which version depends on the version of Cassandra you plan to use. If
-  unsure, use Java 7 as it is known to work with current versions of Cassandra).
+  unsure, use Java 8 as it is known to work with current versions of Cassandra).
```
bschoening, line 50: Perhaps reference Java 11 instead here.  Java 8 is on LTS.

tests/__init__.py
```diff
@@ -8,16 +6,3 @@
 
 if sys.version_info < (3, 0):
```
bschoening, line 7: Python < 3.7 is EOL

README.md
```diff
@@ -7,17 +7,47 @@ localhost.
 The goal of ccm and ccmlib is to make it easy to create, manage and destroy a
 small Cassandra cluster on a local box. It is meant for testing a Cassandra cluster.
 
+New to Python development?
+--------------------------
+Python has moved on since CCM started development. `pip` is the new `easy_install`,
+Python 3 is the new 2.7, and virtualenv is strongly recommended for managing multiple Python versions
```
bschoening, line 13: Suggest: Python 3.7 or later is required and using virtualenv is strongly recommended for managing multiple Python versions

README.md
aweisberg: Good suggestion. I updated the text. Please LMK what you think.

aweisberg:
@bereng @ekaterinadimitrova2 Hi! I updated this PR based on the feedback. Also added a new gotcha to the readme to help people work around AirPlay also listening on port 7k by default.

README.md
ekaterinadimitrova2: I think maybe 8 makes sense until we still maintain as a Community Cassandra versions like 3.x which are on Java 8. I believe we maintain them until next year or next major release.

aweisberg: This comment thread is on an outdated version of the diff. I updated it to reflect that we build on 8 and support 11.

ekaterinadimitrova2: Sounds great! Thank you so much!

bereng:
+1

PR #759, opened 2023-11-06
Remove obsolete use of 'six' for Python 2.7 compatibility

bschoening:


michaelsembwever:
let's come back to this post donation.

i think we still want ccm to work with python2.7 for cqlsh tests and older branches… 
post-donation maybe we should be looking at how we version ccm.

PR #764, opened 2024-04-10
Make a copy so nodes can't mutate each others environment variables

aweisberg:


jmckenzie-dev:
LGTM. Passing curiosity if there's anything else we're assigning to a ref that could later be mutated, but looks like it'd just be the cluster object which... yeah. If we're changing that outside we have bigger problems.

👍

PR #766, opened 2024-05-08
Fix Java version selection

jacek-lewandowski:
Problem: for Cassandra 5 not built from the sources, Java version is not selected properly - it is assumed it works with either current version or Java 8.

Fixes:
- For Cassandra 5.1+ binary distribution, the supported versions are retrieved from cassandra.in.sh
- If supported Cassandra version cannot be retrieved from either build.xml (source dist) or cassandra.in.sh (bin dist), use the hardcoded version - the conditions cover all supported Cassandra versions starting from 2.2.
- Refactored Java selection logic - clearly distinguish supported Java versions by Cassandra and available Java versions in the environment (distributions configured via JAVA_HOME and JAVAn_HOME env variables); prioritizes the supported Java version over the current Java version configured via PATH (which is currently needed for upgrade testing)
- More explanatory messages about the Java version selection
- Added --jvm-version command line option for explicitly setting Java version with cluster or node start commands
- Fixed Cassandra version used to resolve Java version for running node tool (it was using Cassandra version defined for the cluster instead of the node which might have been upgraded)

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -840,7 +840,7 @@ def get_jdk_version(process='java'):
     try:
         version = subprocess.check_output([process, '-version'], stderr=subprocess.STDOUT)
     except OSError:
-        print_("ERROR: Could not find java. Is it in your path?")
+        print_("ERROR: Could not find {}. Is it in your path?".format(process))
         exit(1)
 
     return _get_jdk_version(version)
@@ -931,15 +931,21 @@ def _update_java_version(current_java_version, current_java_home_version,
 
     # Determine "versions" - supported Java versions for select Cassandra distribution
     # First - attempt to get the supported Java versions from the Cassandra distribution
-    build_versions = get_supported_jdk_versions(install_dir)
+    build_versions = None
+    if install_dir:
+        build_versions = get_supported_jdk_versions(install_dir)
     run_versions = build_versions
 
     # If the supported Java versions are not available from the distribution, use the known defaults
+
+    if cassandra_version and not isinstance(cassandra_version, LooseVersion):
+        cassandra_version = LooseVersion(cassandra_version)
+
     if build_versions is None:
-        if cassandra_version >= '5.0':
+        if cassandra_version and cassandra_version >= LooseVersion('5.0'):
             build_versions = [11, 17]
             run_versions = [11, 17]
-        elif cassandra_version >= '4.0':
+        elif cassandra_version and cassandra_version >= LooseVersion('4.0'):
             if not os_env:
                 os_env = os.environ
             if 'CASSANDRA_USE_JDK11' not in os_env:
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -854,38 +854,53 @@ def _get_jdk_version(version):
     ver_pattern = '\"(\d+).*\"'
     return re.search(ver_pattern, str(version)).groups()[0] + ".0"
 
+
+def get_supported_jdk_versions_internal(path, pattern):
+    if os.path.exists(path):
+        with open(path) as f:
+            for line in f:
+                match = re.search(pattern, line)
+                if match:
+                    versions = match.group(1).split(',')
+                    versions = [8 if v == '1.8' else int(v) for v in versions]
+                    return versions
+    return None
+
+
 def get_supported_jdk_versions(install_dir):
     """
-    Return the supported java versions from build.xml
-    Only works in > 4.1
+    Return the supported Java versions from build.xml (source distributions)
+    or from cassandra.in.sh (binary distributions) if such information is present (5.0+ for source, 5.1+ for binary).
     """
-    build = os.path.join(install_dir, 'build.xml')
-    with open(build) as f:
-        for line in f:
-            match = re.search('name="java\.supported" value="([0-9.,]+)"', line)
-            if match:
-                versions = match.group(1).split(',')
-                versions = [8 if v == '1.8' else int(v) for v in versions]
-                return versions
-    raise ValueError("did not find java.supported in build.xml!")
+
+    # source distributions have supported Java versions specified in build.xml since 5.0
+    versions = get_supported_jdk_versions_internal(os.path.join(install_dir, 'build.xml'),
+                                                   'name="java\\.supported" value="([0-9.,]+)"')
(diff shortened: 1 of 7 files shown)


ccmlib/common.py
```diff
@@ -918,22 +918,17 @@ def _update_java_version(current_java_version, current_java_home_version,
...
+    if cassandra_version >= '5.0':
         if os.path.exists(os.path.join(install_dir, 'build.xml')):
             build_versions = get_supported_jdk_versions(install_dir)
             run_versions = get_supported_jdk_versions(install_dir)
         else:
             # binary installs we don't know which jdks are supported any more
-            build_versions = [current_java_version]
-            run_versions = [current_java_version]
+            build_versions = [11, 17]
+            run_versions = [11, 17]
```
michaelsembwever, line 929: this introduces a maintenance point.  every major version that bumps the supported jdks we break here, and have to go fix it.

my understanding is the node will fail-fast (won't start) with the appropriate message, so i'd rather not introduce these two lines.

michaelsembwever:
lgtm, aside from one comment.

this needs to be tested thoroughly with all the different dtest test types (especially the upgrade variants), on all CI systems (as we filter out how we select jdks differently on each).

ccmlib/common.py
jacek-lewandowski: Without this, automatic jvm setup works up to 4.1; with this, automatic jvm works up to 5.x. 

We already decided to hardcode the supported Java versions - see the code for 4.x and previous versions.

This simple change only makes is less problematic to work with Cassandra 5.x.

michaelsembwever: > We already decided to hardcode the supported Java versions - see the code for 4.x and previous versions.
> This simple change only makes is less problematic to work with Cassandra 5.x.


I see.

Is there really no way, for binary versions, that we can detect the supported versions without hardcoding it … ?

```
curl -s https://raw.githubusercontent.com/apache/cassandra/cassandra-5.0-beta1/build.xml | grep 'property\s*name="java.supported"' |sed -ne 's/.*value="\([^"]*\)".*/\1/p'
```

jacek-lewandowski: Binary versions do not have build.xml; I think we should maybe store the supported java versions in some file; but on the other hand - this is really not a big deal to update CCM every couple of years for the supported java versions

michaelsembwever: it can be done without the network too, as `java_versions_supported` is also found in `bin/cassandra.in.sh`
https://github.com/apache/cassandra/blob/trunk/bin/cassandra.in.sh#L117

michaelsembwever:
Thinking more about this… 
```
if path_jvm_version in supported_versions
  use path_jvm_version
if java_home_jvm_version in supported_versions
  use java_home_jvm_version
else
  find JAVA[supported version]_HOME env variable and use that version if env variable found
```

Wouldn't it be a much simpler PR if we just put in an assert that PATH is correct according to JAVA_HOME ? 

What do we gain from adding the complexity that we can honour either PATH or JAVA_HOME ?

ccmlib/common.py
```diff
@@ -854,20 +854,28 @@ def _get_jdk_version(version):
...
-            match = re.search('name="java\.supported" value="([0-9.,]+)"', line)
-            if match:
-                versions = match.group(1).split(',')
-                versions = [8 if v == '1.8' else int(v) for v in versions]
-                return versions
-    raise ValueError("did not find java.supported in build.xml!")
+
+    versions = get_supported_jdk_versions_internal(os.path.join(install_dir, 'build.xml'), 'name="java\.supported" value="([0-9.,]+)"')
+    if versions is None:
+        versions = get_supported_jdk_versions_internal(os.path.join(install_dir, 'bin', 'cassandra.in.sh'), 'java_versions_supported=([0-9.,]+)')
```
michaelsembwever, line 876: nit:  the former works for source ccm versions, the latter for binary versions, checking both seems odd.   but maybe this is as effective as it gets … 🤷

ccmlib/common.py
michaelsembwever: this is fixed now (a few lines above).  bin/cassandra.in.sh is only found in trunk, so 5.0 still needs the hardcode.

ccmlib/common.py
michaelsembwever: i was hoping there was some flag that says whether this is a downloaded or built install.

the conditional fall-throw code is not intuitive.

my suggestion is just add a comment…  
```suggestion
    // source built installs from 5.0 have supported java versions specified in build.xml
    versions = get_supported_jdk_versions_internal(os.path.join(install_dir, 'build.xml'), 'name="java\.supported" value="([0-9.,]+)"')
    if versions is None:
        // downloaded binary installs from 5.1 have supported java versions specified in bin/cassandra.in.sh
        versions = get_supported_jdk_versions_internal(os.path.join(install_dir, 'bin', 'cassandra.in.sh'), 'java_versions_supported=([0-9.,]+)')
```

michaelsembwever (approved):

michaelsembwever (approved):

jacek-lewandowski:
The dtest branch for verification: https://github.com/jacek-lewandowski/cassandra-dtest/commit/13ac8ac168ffb78c26a11d389bda4c3085fba02a

The cassandra branch for verification: https://github.com/jacek-lewandowski/cassandra/commit/0356018e00e41d13147e345b362f269e934b2d14 in CircleCI

jacek-lewandowski:
https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/1354/workflows/82c94bfc-3095-416e-b494-cb956a8989e0/jobs/72770

jacek-lewandowski:
http://34.88.161.37:8080/blue/organizations/jenkins/cassandra/detail/cassandra/65/pipeline/

michaelsembwever:
> http://34.88.161.37:8080/blue/organizations/jenkins/cassandra/detail/cassandra/65/pipeline/

this isn't a permanent or accessible-to-all url.
it's result files are in
[ccm_pull_766_65_ci.zip](https://github.com/riptano/ccm/files/15272680/ccm_pull_766_65_ci.zip)

michaelsembwever:
+1

PR #768, opened 2024-05-15
Cassandra 19636

jacek-lewandowski:
The logic works as follows:

1. Look for available Java versions in the env - figure out versions of Java located at `JAVA_HOME` and all `JAVAx_HOME` env vars (actually execute `java -version` to check)

2. Figure out supported Java versions by Cassandra distribution - for Cassandra 5.1+ binary distribution the supported Java versions can be read from `cassandra.in.sh` and for Cassandra 5.0+ source distribution the supported Java versions can be read from `build.xml`. For other Cassandra versions, the supported Java versions are hardcoded in CCM

3. If Java version is explicitly provided - use it if it is available (see 1), warn if it is not supported by Cassandra

4. If `JAVA_HOME` is defined, use that Java distribution if it is supported by Cassandra distribution; if it is not supported, select the closest supported (see 2) and available (see 1) Java version

5. If `JAVA_HOME` is not defined, use the highest supported (see 2) and available (see 1) Java version

Also, if Java is available on the `PATH`, `JAVA_HOME` must be defined as well and both must refer to the same Java version. Fail if Java is available on the `PATH` and there is no `JAVA_HOME` or if `JAVA_HOME` refers a different Java version than Java defined on the `PATH`.

Also,
- Added `--jvm-version` option to the cluster and node start commands.
- Fixed one problem with Java selection for tools run on the upgraded node

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -825,7 +825,10 @@ def invalidate_cache():
 
 
 def get_jdk_version_int(process='java'):
-    jdk_version = float(get_jdk_version(process))
+    jdk_version_str = get_jdk_version(process)
+    if jdk_version_str is None:
+        return None
+    jdk_version = float(jdk_version_str)
     # Make it Java 8 instead of 1.8 (or 7 instead of 1.7)
     jdk_version = int(jdk_version if jdk_version >= 2 else 10 * (jdk_version - 1))
     return jdk_version
@@ -840,8 +843,8 @@ def get_jdk_version(process='java'):
     try:
         version = subprocess.check_output([process, '-version'], stderr=subprocess.STDOUT)
     except OSError:
-        print_("ERROR: Could not find java. Is it in your path?")
-        exit(1)
+        info("Could not find {}.".format(process))
+        return None
 
     return _get_jdk_version(version)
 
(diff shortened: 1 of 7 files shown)


ccmlib/common.py
```diff
@@ -927,18 +927,26 @@ def get_available_jdk_versions(env):
...
                         for_build=False, info_message=None):
     """
-    Selects Java distribution and returns the updated env (updates JAVA_HOME and PATH variables if needed).
-    If 'jvm_version' is explicitly set, that one will be used if a Java distribution of that version can be found.
-    Otherwise, the Java version will be guessed from the distribution provided by 'install_dir' parameter or from
-    the provided 'cassandra_version' parameter.
-    :param jvm_version: The explicit Java version to use - must be the major Java version number like 8 or 11 (int).
+    Updates or fixes the environment variables for Java distribution.
+    If 'jvm_version' is explicitly set, that one will be used as long as it can be found in the current environment
+    via JAVA_HOME and JAVAx_HOME variables.
```
michaelsembwever, line 932:
```suggestion
    If 'jvm_version' is explicitly set, that one will be used as long as it can be found in the current environment via JAVA_HOME and JAVAx_HOME variables.
```

ccmlib/common.py
```diff
@@ -927,18 +927,26 @@ def get_available_jdk_versions(env):
...
+    Otherwise, the supported Java versions will be guessed from the provided `cassandra_version` (or install dir if None).
```
michaelsembwever, line 933:
```suggestion
    Otherwise, the supported Java versions are used from the provided `cassandra_version` (or install dir if None).
```

ccmlib/common.py
```diff
@@ -927,18 +927,26 @@ def get_available_jdk_versions(env):
...
+    Then, if there is a JAVA_HOME variable set, that Java distribution will be used if it is supported by Cassandra.
+    If not, the closest supported Java version will be used from the available Java distributions (or the highest
+    supported version if JAVA_HOME is not defined).
```
michaelsembwever, line 936:
```suggestion
    If not, the closest supported Java version will be used from the available Java distributions (or the highest supported version if JAVA_HOME is not defined).
```

ccmlib/common.py
michaelsembwever, line 932: Yes.  This is important.  Folk will want to override a C* versions supported jdk.  In both source and binary releases this is now possible in C*, intended only for development reasons.

ccmlib/common.py
```diff
@@ -986,40 +1000,36 @@ def _update_java_version(current_java_version, current_java_home_version,
...
-                        .format(info_message, current_java_home_version, cassandra_version, supported_versions))
-
-            # nothing more to be done, just add JAVA_HOME/bin to the PATH
+        elif current_java_home_version and current_java_home_version in supported_versions:
+            info('{}: Using the current Java {} available from JAVA_HOME for the current invocation of Cassandra {}.'
+                 .format(info_message, current_java_version, cassandra_version))
+            # just need to add JAVA_HOME/bin to the PATH
             return update_path_in_env(env, env['JAVA_HOME'] + '/bin')
 
-        # Fallback - use the highest supported Java version available in the current env
```
michaelsembwever, line 1012: nit: the default jdk for any C* version is always its lowest jdk 
🤷

ccmlib/common.py
```diff
@@ -986,40 +1000,36 @@ def _update_java_version(current_java_version, current_java_home_version,
...
+                    .format(info_message, current_java_version, cassandra_version, supported_versions))
+        else:
+            warning('{}: JAVA_HOME is not defined; CCM will try to find a suitable Java distribution among JAVAx_HOME env variables.')
+
+        if len(supported_available_versions) > 0:
+            info('{}: CCM has found {} Java distributions, the required Java version for Cassandra {} is {}.'.format(
+                info_message, str(available_versions), cassandra_version, supported_versions))
+            ref_version = current_java_home_version if current_java_home_version else max(supported_available_versions)
+            # sort by the absolute difference between the current java version and the supported version
+            # to use the closest version if the current version is not supported
```
michaelsembwever, line 1026: nit: the default jdk for any C* version is always its lowest jdk
🤷

(the rationale here to be different is that we're assuming the closest will involve less jdk-switching over possible upgrade steps (the caller is doing, e.g. dtest-upgrade-large) and therefore enable more such steps.  i'm not entirely sure this is the best approach, it might be more intuitive that C* versions always fall back to preferring their default jdk…)

michaelsembwever (approved):

michaelsembwever:
CI results attached to jira ticket.

PR #769, opened 2024-05-21
Fix interpretation of CASSANDRA_USE_JDK11 and auto set CASSANDRA_USE_JDK11 when needed

jacek-lewandowski:


ccmlib/common.py
```diff
@@ -903,17 +906,21 @@ def get_supported_jdk_versions(cassandra_version, install_dir, for_build, env):
         if cassandra_version and cassandra_version >= LooseVersion('5.0'):
             build_versions = [11, 17]
             run_versions = [11, 17]
+            info("Cassandra 5.0+ detected, using Java 11 and 17")
         elif cassandra_version and cassandra_version >= LooseVersion('4.0'):
-            if 'CASSANDRA_USE_JDK11' not in env:
+            if 'CASSANDRA_USE_JDK11' not in env or env['CASSANDRA_USE_JDK11'] != 'true':
```
michaelsembwever, line 911:
```suggestion
            if 'CASSANDRA_USE_JDK11' not in env or env['CASSANDRA_USE_JDK11'].lower() != 'true':
```

michaelsembwever: build.xml (ant) accepts all of `true`, `yes`, and `on`, case-insensitive…

ccmlib/common.py
```diff
@@ -971,6 +978,14 @@ def update_path_in_env(env, path_entry):
     return env
 
 
+def _maybe_set_use_jdk11_env(env, jvm_version, cassandra_version):
+    if jvm_version >= 11 and cassandra_version and '4.0' <= cassandra_version < '5.0':
+        env['CASSANDRA_USE_JDK11'] = 'true'
+    elif 'CASSANDRA_USE_JDK11' in env:
+        del env['CASSANDRA_USE_JDK11']
+    return env
```
michaelsembwever, line 986: do we really want to introduce this ? given everything is so brittle and every change seems to be breaking _something_

otherwise i agree with the intent:  the caller may be specifying a binary version, thinking CASSANDRA_USE_JDK11 is never needed, but ccm can't download it and fallbacks to building the source of it…

ccmlib/common.py
jacek-lewandowski: as we talked, instead of doing this, I'm removing the interpretation of this env variable

jacek-lewandowski:
(4.0-j8) https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/1370/workflows/3e2deef9-2f52-4399-8166-2d82617dce5e

ccmlib/common.py
```diff
@@ -908,14 +908,9 @@ def get_supported_jdk_versions(cassandra_version, install_dir, for_build, env):
...
-                build_versions = [8, 11]
-                run_versions = [8, 11]
-                info("Cassandra 4.0+ detected, using Java 8 and 11")
-            else:
-                build_versions = [11]
-                run_versions = [11]
-                info("Cassandra 4.0+ detected, using Java 11")
+            build_versions = [8, 11]
+            run_versions = [8, 11]
+            info("Cassandra 4.0+ detected, using Java 8 and 11")
```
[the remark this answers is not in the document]
michaelsembwever: oh, when we build ccm it sets the `-Duse.jdk11=true` flag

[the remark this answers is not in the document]
michaelsembwever: so, will this still break the following situation…?

The env is path+java_home 8, java11_home is set, CASSANDRA_USE_JDK11 is true, ccm source 4.0

ccm goes to build cassandra-4.0 using jdk8.  it fails because CASSANDRA_USE_JDK11 is true.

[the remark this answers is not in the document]
michaelsembwever: the existing contract/expectation here was that `CASSANDRA_USE_JDK11=true` mean switch to use jdk11

michaelsembwever (approved):

michaelsembwever:
I've got a few CI jobs running, will post their results asap

ccmlib/common.py
```diff
@@ -899,21 +902,26 @@ def get_supported_jdk_versions(cassandra_version, install_dir, for_build, env):
...
+            info("Cassandra 5.0+ detected, using Java 11 and 17")
         elif cassandra_version and cassandra_version >= LooseVersion('4.0'):
-            if 'CASSANDRA_USE_JDK11' not in env:
-                build_versions = [8, 11]
-                run_versions = [8, 11]
-            else:
+            if isCassandraUseJDK11Set:
                 build_versions = [11]
                 run_versions = [11]
+                info("Cassandra 4.0+ detected, using Java 11 (Java 8 is excluded because CASSANDRA_USE_JDK11 is set)")
```
michaelsembwever, line 915:
```suggestion
                info("Cassandra 4.0+ detected, using Java 11 (Java 8 is excluded because the deprecated CASSANDRA_USE_JDK11 is set)")
```

PR #770, opened 2024-05-22
Fix confused env usages

jacek-lewandowski:
1. Before/after starting a node, only JAVA_HOME was stored/restored which could be responsible for divergences between Java version on PATH and JAVA_HOME
2. Method which calls 'java' from the PATH to check its version didn't take into account the updated env


michaelsembwever (approved):

aweisberg (approved):
+1 TY for fixing this. I checked and it looks like all the tests pass and it may have actually fixed some pre-existing test failures.

PR #771, opened 2024-05-25
Add github actions CI

jacek-lewandowski:


.github/workflows/main.yml
```diff
@@ -0,0 +1,42 @@
...
+      - uses: actions/checkout@v4
+
+      - name: Create virtual environment
+        run: python -m venv venv
+
+      - name: Activate virtual environment
+        run: source venv/bin/activate
+
+      - name: Install dependencies
+        run: pip install -r requirements.txt && pip install -r tests/requirements.txt
```
Mmuzaf, line 25: https://github.com/riptano/ccm/blob/master/.travis.yml#L3

Travis has different versions, should we use different ones as well?

jacek-lewandowski: where do we run Travis? GitHub actions uses the newest I think

michaelsembwever: we should converged, or remove, the travis stuff

michaelsembwever (approved):
optimally the travis config should be brought into line to match, but is optional IMHO as part of this PR

jacek-lewandowski:
I have removed Travis configuration. It is not used in this repository.

PR #778, opened 2024-09-30
Use git clone bare instead of mirror

michaelsembwever:
mirror with refs explicit isn't supported in newer git version.  bare is want is desired, that is don't get upstream remotes.

 patch by Mick Semb Wever; reviewed by Brandon Williams


michaelsembwever:
previously it was hitting this error: https://apple.stackexchange.com/questions/352404/git-clone-fails-with-fatal-multiple-updates-for-ref-refs-tags-v1-0-0-not-all

PR #787, opened 2025-03-30
CASSANDRA-20328 Also copy tools/bin/cassandra.in.sh for tools

tolbertam:
After [CASSANDRA-20328](https://issues.apache.org/jira/browse/CASSANDRA-20328), sstableloader has been updated to be a separate tool residing in the tools directory.

Existing dtests using sstableloader rely on make_cassandra_env which copies bin/cassandra.in.sh to the node's directory. As tools/bin/sstableloader now relies on tools/bin/cassandra.in.sh, updates this function to also copy this file to the node's path.

Note: There is some code that implies sstableloader should still be ran from `bin/sstableloader`.  I think we should keep that as we want to retain compatibility with previous C* versions, and the new changes in CASSANDRA-20328 still retain backwards compatibility with that path.

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -42,6 +42,7 @@ from ccmlib import extension
 
 
 BIN_DIR = "bin"
+TOOLS_BIN_DIR = "tools/bin"
 CASSANDRA_CONF_DIR = "conf"
 
 CASSANDRA_CONF = "cassandra.yaml"
@@ -394,14 +395,18 @@ def rmdirs(path):
 
 def make_cassandra_env(install_dir, node_path, update_conf=True):
     version_from_build = extension.get_cluster_class(install_dir).getNodeClass().get_version_from_build(node_path=node_path)
-    if is_win() and version_from_build >= '2.1':
-        sh_file = os.path.join(CASSANDRA_CONF_DIR, CASSANDRA_WIN_ENV)
-    else:
-        sh_file = os.path.join(BIN_DIR, CASSANDRA_SH)
-    orig = os.path.join(install_dir, sh_file)
-    dst = os.path.join(node_path, sh_file)
-    if not is_win() or not os.path.exists(dst):
-        shutil.copy(orig, dst)
+    for sh_file_path in [ BIN_DIR, TOOLS_BIN_DIR ]:
+        if is_win() and version_from_build >= '2.1':
+            sh_file = os.path.join(CASSANDRA_CONF_DIR, CASSANDRA_WIN_ENV)
+        else:
+            sh_file = os.path.join(sh_file_path, CASSANDRA_SH)
+
+        orig = os.path.join(install_dir, sh_file)
+        if os.path.exists(orig):
+            dst = os.path.join(node_path, sh_file)
+            if not is_win() or not os.path.exists(dst):
+                os.makedirs(os.path.dirname(dst), exist_ok=True)
+                shutil.copy(orig, dst)
 
     if update_conf and not (is_win() and version_from_build >= '2.1'):
         replacements = [


michaelsembwever:
LGTM.

CI, pre-commit w/ upgrade: 
[ci_summary_ccm_tolbertam-CASSANDRA-20328_203.html.gz](https://github.com/user-attachments/files/19632855/ci_summary_ccm_tolbertam-CASSANDRA-20328_203.html.gz) and 
[results_details_ccm_tolbertam-CASSANDRA-20328_203.tar.xz.gz](https://github.com/user-attachments/files/19632858/results_details_ccm_tolbertam-CASSANDRA-20328_203.tar.xz.gz)

( via https://github.com/apache/cassandra-dtest/compare/trunk...thelastpickle:cassandra-dtest:mck/20328 )

michaelsembwever (approved):

ccmlib/common.py
```diff
@@ -394,14 +395,18 @@ def rmdirs(path):
...
-    if is_win() and version_from_build >= '2.1':
-        sh_file = os.path.join(CASSANDRA_CONF_DIR, CASSANDRA_WIN_ENV)
-    else:
-        sh_file = os.path.join(BIN_DIR, CASSANDRA_SH)
-    orig = os.path.join(install_dir, sh_file)
-    dst = os.path.join(node_path, sh_file)
-    if not is_win() or not os.path.exists(dst):
-        shutil.copy(orig, dst)
+    sh_file_paths = [ BIN_DIR, TOOLS_BIN_DIR ]
+    for sh_file_path in [ BIN_DIR, TOOLS_BIN_DIR ]:
```
smiklosovic, line 398: @tolbertam should not you reuse `sh_file_paths` in this `in`?

tolbertam: ah right, a bit of an oversight, i'll fix it :+1:

tolbertam: removed redundant variable

smiklosovic:
@tolbertam could you check why the check failed?

tolbertam:
Looks like this smoke test is fairly new, but it appears what happened is on first attempt to start, there was a timeout waiting for the pid file to be created (looks like timeout is 30 seconds)

```
+ ./ccm start
/home/runner/work/cassandra-ccm/cassandra-ccm/./ccm:23: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
  import pkg_resources
20:37:42,740 ccm INFO Supported Java versions could not be retrieved from Cassandra distribution in '/home/runner/.ccm/repository/5.0.3'
20:37:42,740 ccm INFO Cassandra 5.0+ detected, using Java 11 and 17
20:38:01,581 ccm INFO node1: Using the current Java 17 available on PATH for the current invocation of Cassandra 5.0.3.
20:38:01,635 ccm INFO Supported Java versions could not be retrieved from Cassandra distribution in '/home/runner/.ccm/repository/5.0.3'
20:38:01,635 ccm INFO Cassandra 5.0+ detected, using Java 11 and 17
20:38:01,844 ccm INFO node1: Using the current Java 17 available on PATH for the current invocation of Cassandra 5.0.3.
20:38:01,872 ccm INFO Starting node1 with JAVA_HOME=/usr/lib/jvm/temurin-17-jdk-amd64 java_version=17 cassandra_version=5.0.3, install_dir=/home/runner/.ccm/repository/5.0.3
20:38:32,442 ccm ERROR Timed out waiting for pidfile to be filled (current time is 2025-04-13 20:38:32.442425, file exists False)
Problem starting node node1 due to [Errno 2] No such file or directory: '/home/runner/.ccm/test/node1/cassandra.pid'
```

The test then continued to retry, and at some point i'm guessing the node fully started, and subsequent requests repeatedly failed with:

```
ccmlib.common.UnavailableSocketError: Inet address 127.0.0.1:7000 is not available: [Errno 98] Address already in use; a cluster may already be running or you may need to add the loopback alias
```

Perhaps we hit a slow build agent?  It looks like the timeout on waiting for the pidfile to be created is 30 seconds (https://github.com/apache/cassandra-ccm/blob/trunk/ccmlib/node.py#L2154).  You can see in the timestamps above that it logs timestamps spanning 50 seconds.

I created a cluster in the same way locally (`ccm create test -v 5.0.3. -n1 --vnodes`) which logs the same statements within the space of ~200ms and the pidfile was always created in < 1 second, odd.

I suspect if I were to run this again that it would succeed...going to give it another run to see if the same thing happens again.

tolbertam:
Kicking off the build again, it ran quite swiftly:

```
00:14:51,481 ccm INFO Supported Java versions could not be retrieved from Cassandra distribution in '/home/runner/.ccm/repository/5.0.3'
00:14:51,481 ccm INFO Cassandra 5.0+ detected, using Java 11 and 17
00:14:51,888 ccm INFO node1: Using the current Java 17 available on PATH for the current invocation of Cassandra 5.0.3.
00:14:51,943 ccm INFO Supported Java versions could not be retrieved from Cassandra distribution in '/home/runner/.ccm/repository/5.0.3'
00:14:51,943 ccm INFO Cassandra 5.0+ detected, using Java 11 and 17
00:14:52,137 ccm INFO node1: Using the current Java 17 available on PATH for the current invocation of Cassandra 5.0.3.
00:14:52,167 ccm INFO Starting node1 with JAVA_HOME=/usr/lib/jvm/temurin-17-jdk-amd64 java_version=17 cassandra_version=5.0.3, install_dir=/home/runner/.ccm/repository/5.0.3
```

It looks like the smoke test does the following for each configuration:
https://github.com/apache/cassandra-ccm/actions/runs/14433125081/workflow?pr=787#L54

```
          ccm_test() { for i in {1..9}; do ./ccm start && ./ccm remove && return 0 || echo retrying; sleep 20; done; echo "ccm start failed after 9 attempts"; exit 1; }
```

I've wrestled with this before, but `ccm remove` may not stop nodes if doesn't detect the nodes as running, which makes sense if the pid file was never created.  It leaves you in a state where you have to kill processes by hand for them to truly go away.  Maybe we can add a `killall java` or something after `./ccm remove` in this script to get around this possibly?   It may make things slightly more resilient to a slow worker.

smiklosovic:
@tolbertam yeah I am all in for making this more robust.

smiklosovic (approved):

michaelsembwever:
Sorry, i didn't see these messages^.

Yes, the smoke test is very new and rudimentary. 

And I've already cleaned it up in this commit that's proposed for inclusion in #790 
https://github.com/thelastpickle/cassandra-ccm/commit/459b14db9903ca93bbbd81df3c9e0a75e1d0ff50 

The idea is just
- use `ccm stop` in between each iteration to ensure all eventually started nodes shutdown again,
- block each iteration until the port is free: `while nc -z 127.0.0.1 7000 ; do echo . ; sleep 1 ; done`

michaelsembwever:
@tolbertam , will you squash your commits and merge ?  (after manual squashing and a force push to this branch, i'm ok if you `rebase and merge` this PR, as this is not the main in-tree cassandra repository.  but the manual trunk push approach is fine too ofc)

tolbertam:
ah yes sorry I lost sight of this myself, thanks for following up!  I'll squash and fix up the commit message and push on this branch as a single commit, assuming it passes CI i'll rebase and merge :+1:

tolbertam:
First two runs seemed to get in a  `ccm stop` loop but it ran swiftly the third time, suspect maybe something still to do with the pidfile there.  Going to go ahead and merge as I don't think the changes on the PR are contributing :+1:

PR #790, opened 2025-05-02
Fix panic when paramiko is installed

dkropachev:
When `paramiko` is installed CCM is adding remote part of CLI. In order to build it, CCM reads renders `argparser` usage string. Recently `argparser` have changed format and now `optional arguments:` is not there anymore.
Fix code to not to fail in any case.

Failure:
```
ccm create gocql_cassandra_integration_test -i 127.0.2 -v 4.1.6 -n 3 -d --vnodes --jvm_arg="-Xmx256m -XX:NewSize=100m"

  File "/home/dmitry.kropachev/.local/bin/ccm", line 107, in <module>
    parser = cmd.get_parser()
  File "/home/dmitry.kropachev/.local/lib/python3.10/site-packages/ccmlib/cmds/command.py", line 60, in get_parser
    get_remote_usage()
  File "/home/dmitry.kropachev/.local/lib/python3.10/site-packages/ccmlib/remote.py", line 32, in get_remote_usage
    return RemoteOptionsParser().usage()
  File "/home/dmitry.kropachev/.local/lib/python3.10/site-packages/ccmlib/remote.py", line 497, in usage
    usage = self.parser.format_help().split("optional arguments:")[1]
IndexError: list index out of range
```

### DIFF
--- a/ccm
+++ b/ccm
@@ -24,20 +24,10 @@ import pkg_resources
 from six import print_
 
 from ccmlib import common
-from ccmlib.cmds import cluster_cmds, command, node_cmds
+from ccmlib.cmds import cluster_cmds, node_cmds
+from ccmlib.cmds.common import get_command
 from ccmlib.remote import (PARAMIKO_IS_AVAILABLE, execute_ccm_remotely,
                            get_remote_options, get_remote_usage)
-
-
-def get_command(kind, cmd):
-    cmd_name = kind.lower().capitalize() + cmd.lower().capitalize() + "Cmd"
-    try:
-        klass = (cluster_cmds if kind.lower() == 'cluster' else node_cmds).__dict__[cmd_name]
-    except KeyError:
-        return None
-    if not issubclass(klass, command.Cmd):
-        return None
-    return klass()
 
 
 def print_subcommand_usage(kind):
--- a/ccmlib/cmds/command.py
+++ b/ccmlib/cmds/command.py
@@ -25,6 +25,7 @@ import re
 
 from six import print_
 
+import ccmlib
 from ccmlib import common
 from ccmlib.cluster_factory import ClusterFactory
 from ccmlib.remote import PARAMIKO_IS_AVAILABLE, get_remote_usage
@@ -70,7 +71,8 @@ class Cmd(object):
     def get_parser(self):
         if self.usage == "":
             pass
-        if PARAMIKO_IS_AVAILABLE:
+        # Do not collapse to PARAMIKO_IS_AVAILABLE, we need to pull it dynamically for testing purposes
+        if ccmlib.remote.PARAMIKO_IS_AVAILABLE:
             self.usage = self.usage.replace("usage: ccm",
                                             "usage: ccm [remote_options]") + \
                          os.linesep + os.linesep + \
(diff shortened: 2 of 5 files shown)


dkropachev:
@michaelsembwever please take a look.

michaelsembwever:
any suggestions for a test ?

dkropachev:
> any suggestions for a test ?

```
pip install paramiko
ccm ...
```

michaelsembwever:
was thinking for under `tests/`  ,   we're very low on coverage.

we could add 
```
pip install paramiko
ccm --help
```
to `.github/workflows/main.yml` 
but that feels like overkill compared to adding a test under `test/` 🤷

michaelsembwever:
small suggestions to add here @dkropachev https://github.com/apache/cassandra-ccm/commit/459b14db9903ca93bbbd81df3c9e0a75e1d0ff50

simple test added to gha.  clean up usage text. fix "is not" against string literal.

https://github.com/thelastpickle/cassandra-ccm/actions/runs/14824735976

dkropachev:
> small suggestions to add here @dkropachev [459b14d](https://github.com/apache/cassandra-ccm/commit/459b14db9903ca93bbbd81df3c9e0a75e1d0ff50)
> 
> simple test added to gha. clean up usage text. fix "is not" against string literal.
> 
> https://github.com/thelastpickle/cassandra-ccm/actions/runs/14824735976

Thanks, added test, please take a look.

ccm
```diff
@@ -24,22 +24,12 @@ import pkg_resources
...
 
 from ccmlib import common
-from ccmlib.cmds import cluster_cmds, command, node_cmds
+from ccmlib.cmds import cluster_cmds, node_cmds
+from ccmlib.cmds.common import get_command
 from ccmlib.remote import (PARAMIKO_IS_AVAILABLE, execute_ccm_remotely,
                            get_remote_options, get_remote_usage)
 
 
-def get_command(kind, cmd):
```
michaelsembwever, line 32: does this break compatibility ?  would any client be calling this ? (any client that cannot easily add `from ccmlib.cmds.common import get_command` ?) 

if so, we should put in (something like) a deprecated
```
def get_command(kind, cmd):
    import ccmlib.cmds.common
    return common.get_command(kind, cmd)
```

(`common.get_version_from_build` has a deprecated coding example, not sure what's idiomatic here)

dkropachev: I don't think so:
-  `ccm` is executable` no one should use it, when you import a method it becomes and there is no way
- when you import anything it becomes part of the package and you can import it as well.
Example:
ccmlib.cmds.command.py imports `PARAMIKO_IS_AVAILABLE` from `ccmlib.remote`
But you can import `PARAMIKO_IS_AVAILABLE` from `ccmlib.cmds.command` too:
```
from ccmlib.cmds.command import PARAMIKO_IS_AVAILABLE
```

michaelsembwever: if folk cannot import `ccm.get_command` then we're good.

dkropachev: I have tried and failed, don't think you can import anything from the root.

michaelsembwever (approved):

PR #795, opened 2025-09-18
When creating a cluster and validating the jdk, check the cassandra version

michaelsembwever:
`common.assert_jdk_valid_for_cassandra_version` performs the jdk version check based on only cassandra versions

### DIFF
--- a/ccmlib/cmds/cluster_cmds.py
+++ b/ccmlib/cmds/cluster_cmds.py
@@ -128,9 +128,10 @@ class ClusterCreateCmd(Cmd):
                 common.validate_install_dir(options.install_dir)
             except ArgumentError:
                 parser.print_help()
-                parser.error("%s is not a valid cassandra directory. You must define a cassandra dir or version." % options.install_dir)
+                parser.error("%s is not a valid cassandra directory. You must define install-dir or version." % options.install_dir)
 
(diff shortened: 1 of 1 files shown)


bereng:
+1

szymon-miezal (approved):
The patch makes sense to me, it allows to rely on single source of truth.
DSE 6.x can run on both Java 8 and 11, AFAIR it also adversities itself as C* 4.x, which means running on both versions will be possible.

ccmlib/cmds/cluster_cmds.py
```diff
@@ -128,9 +128,10 @@ def validate(self, parser, options, args):
                 common.validate_install_dir(options.install_dir)
             except ArgumentError:
                 parser.print_help()
-                parser.error("%s is not a valid cassandra directory. You must define a cassandra dir or version." % options.install_dir)
+                parser.error("%s is not a valid cassandra directory. You must define install-dir or version." % options.install_dir)
 
-            common.assert_jdk_valid_for_cassandra_version(extension.get_cluster_class(options.install_dir).getNodeClass().get_version_from_build(options.install_dir))
+            node_class = extension.get_cluster_class(options.install_dir).getNodeClass()
+            common.assert_jdk_valid_for_cassandra_version(node_class.get_version_from_build(options.install_dir, cassandra=True))
```
michaelsembwever, line 134: adding `cassandra=True` is the fix here.

PR #803, opened 2026-03-12
Fix parsing of java_versions_supported in bin/cassandra.in.sh

michaelsembwever:
patch by Mick Semb Wever; reviewed by xxx for CASSANDRA-18831

### DIFF
--- a/.github/workflows/main-python-2-7.yml
+++ b/.github/workflows/main-python-2-7.yml
@@ -20,6 +20,8 @@ on:
   push:
   pull_request:
   workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * 0'
 
 jobs:
   lint_test_smoke:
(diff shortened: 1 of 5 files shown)


.github/workflows/main-python-2-7.yml
```diff
@@ -20,6 +20,8 @@ on:
   push:
   pull_request:
   workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * 0'
```
michaelsembwever, line 24: adding weekly build of default (trunk) branch, so we catch such failures a bit faster next time

michaelsembwever:
still hitting an issue with the smoke test on 4.0.19 using python2.17

michaelsembwever:
ready for review @driftx

driftx (approved):
Ticket numbers seems to wrong, but lgtm.

michaelsembwever:
> Ticket numbers seems to wrong,  

the jdk21 patch broke the parsing of parsing of `java_versions_supported` in `bin/cassandra.in.sh`

and the fix (primarily) in `ccmlib/common.py` is so small and trivial, it made sense to just use the existing ticket 🤷

michaelsembwever:
merged with https://github.com/apache/cassandra-ccm/commit/dfa4e82ba46346cd87a85f784c66b796a6334e7b

